PR #3, opened 2015-10-08
Add managing procedures CIP

boggle:
This proposal introduces stored procedures as a new concept to Cypher and proposes the addition of various facilities
for calling and managing them.

User-defined functions are a separate concern that might be addressed in a future CIP.


boggle:
Closing this; will be superseded by new PR on calling procs only.

PR #8, opened 2015-10-13
Added the EXISTS CIP

systay:
Link to a viewable CIP: https://github.com/systay/cypher/blob/69725d2e61b73729705cab483e3012c2df845861/cip/CIP2015-05-13-EXISTS.adoc

PR #10, opened 2015-10-27
Adds the EXISTS CIP

systay:
Replaces #8

### DIFF
--- a/cip/CIP2015-05-13-EXISTS.adoc
+++ b/cip/CIP2015-05-13-EXISTS.adoc
@@ -22,7 +22,7 @@ This form is helpful, but not exhaustive in the types of subqueries that users w
 
 == Proposal
 
-To make this feature more powerful, add a new EXISTS-function, that would exist in two forms:
+To make this feature more powerful, add a new function named +EXISTS+, that would exist in two forms:
 
  * Property existence checking
  * Subquery existence checking
@@ -51,15 +51,14 @@ Note: the read-only clause above includes WITH but not RETURN.
 === Semantics
 
 All forms of +EXIST(...)+ accomplish the same task - check if a particular pattern exists in the database.
-They are expressions, and as such must be side-effect free. That is, the subqueries in +EXIST+ must not be updating queries.
+They are expressions, and as such must be side-effect free; that is, the subqueries in +EXIST+ must not be updating queries.
 Regarding nomenclature, all forms of +EXISTS+ are scalar expressions - they return a single boolean value, and this value is never +null+.
 
 ==== Property exists
 
 In the property checking form, any expression is accepted as the input.
-If the input expression returns a +null+ value, +EXIST(...)+ returns false, and for all other values, it will return +true+.
-This form of +EXISTS(...)+ is equivalent to checking for +null+ using +IS NOT NULL+.
-I.e. +EXISTS(expr)+ is equivalent to +expr IS NOT NULL+.
+If the input expression returns a +null+ value, +EXISTS(...)+ returns false, and for all other values, it will return +true+.
+This form of +EXISTS(...)+ is equivalent to checking for +null+ using +IS NOT NULL+; i.e. +EXISTS(expr)+ is equivalent to +expr IS NOT NULL+.
 
 This form is there to make property existence checking idiomatic in Cypher - we do not want the user thinking of missing properties as checking for +null+ values.
 
--- a/cip/CIP2015-05-13-EXISTS.adoc
+++ b/cip/CIP2015-05-13-EXISTS.adoc
@@ -17,13 +17,15 @@ toc::[]
 
 == Background
 
-In a way, Cypher already has existential subqueries - pattern predicates allow checking for the existence of subgraphs. This form is helpful, but not exhaustive in the forms of subqueries that users want to be able to express.
+In a way, Cypher already has existential subqueries - pattern predicates allow checking for the existence of subgraphs.
+This form is helpful, but not exhaustive in the types of subqueries that users want to be able to express.
 
 == Proposal
 
 To make this feature more powerful, add a new EXISTS-function, that would exist in two forms:
- . Property existence checking
- . Subquery existence checking
+
+ * Property existence checking
+ * Subquery existence checking
 
 === Syntax
 
@@ -34,8 +36,7 @@ expression = …
            | simple subquery exists
             ;
 
-property exists = "EXISTS", "(", expression, ".", identifier ")"
-                | "EXISTS", "(", expression, "[", expression, "]", ")"
+property exists = "EXISTS", "(", expression ")"
                 ;
 
 subquery exists = "EXISTS", "{", read only clause, { read only clause }, "}" ;
@@ -51,22 +52,34 @@ Note: the read-only clause above includes WITH but not RETURN.
 
 All forms of +EXIST(...)+ accomplish the same task - check if a particular pattern exists in the database.
 They are expressions, and as such must be side-effect free. That is, the subqueries in +EXIST+ must not be updating queries.
+Regarding nomenclature, all forms of +EXISTS+ are scalar expressions - they return a single boolean value, and this value is never +null+.
 
 ==== Property exists
 
-In the property checking form, either an identifier or a string expression is used to represent the property key name. If the node/relationship returned has a property with the given name, the +EXISTS(...)+ expression should return true, otherwise it should return false. If the node/relationship expression returns a null value, the +EXISTS(...)+ expression should return null as well.
+In the property checking form, any expression is accepted as the input.
+If the input expression returns a +null+ value, +EXIST(...)+ returns false, and for all other values, it will return +true+.
+This form of +EXISTS(...)+ is equivalent to checking for +null+ using +IS NOT NULL+.
+I.e. +EXISTS(expr)+ is equivalent to +expr IS NOT NULL+.
 
-This form of +EXISTS+ replaces the pre-existing +has(...)+ function.
+This form is there to make property existence checking idiomatic in Cypher - we do not want the user thinking of missing properties as checking for +null+ values.
 
 ==== Subquery exists
-Identifiers and parameters from the outer query are available in the subquery. Identifiers introduced in the subquery are not available on the outside.
+When checking for the existence of subqueries, variables and parameters from the outer query are available in the subquery.
+Variables introduced in the subquery are not available on the outside.
 
-For each matching subgraph evaluated with +EXISTS(...)+, the result value must be true if the subquery finds at least one matching row. If no matches are found, false should be returned.
+For each matching subgraph evaluated with +EXISTS{...}+, the result value must be true if the subquery finds at least one matching row.
+If no matches are found, false should be returned.
 
-Regarding nomenclature, +EXISTS(...)+ is a scalar expression - it returns a single boolean value.
+The subquery used in +EXISTS{...}+ has to follow the following rules:
+
+ * It has to be a read-only query. Updating the graph as part of a predicate is not allowed.
+ * It cannot end in a +RETURN+ clause, nor may it contain +UNION+
+ * If +WITH+ ends the subquery, it should be able to change the cardinality of the subquery in an for +EXISTS(...)+ interesting way.
+In other words, they should include +WHERE+ or +SKIP+, since they could turn a query that returns rows into one that does not.
 
 ==== Simple subquery exists
 When the subquery can be described with a single +MATCH-WHERE+ clause, the +MATCH+ keyword can be omitted.
+The difference between this form and a simple pattern predicate, which is already available is that this form allows for introducing new variables inside the +EXISTS(...)+ scope.
 
 === Examples
 _Example 1A:_
@@ -80,13 +93,13 @@ RETURN actor
 ----
 
 _Example 1B:_
+Return all nodes with both first name and second name:
 [source, cypher]
 ----
 MATCH (actor)
-WHERE EXISTS(actor["slogan"])
+WHERE EXISTS(actor.firstName + actor.lastName)
 RETURN actor
 ----
-
 
 _Example 2A:_
 
@@ -104,7 +117,7 @@ RETURN actor
 
 _Example 2B:_
 
-Find all actors who have acted together with another actor with the same name.  +
+Find all actors who have acted together with another actor with the same name.
 
 [source, cypher]
 ----
@@ -117,18 +130,18 @@ RETURN actor
 ----
 
 === Interaction with existing features
-This replaces the pre-existing +has(...)+ function.
+This replaces the pre-existing +HAS(...)+ function.
 
 == What others do
 
 This is very similar to what SQL does with its EXISTS functionality.
 
-This is also very similar in syntax to what SPARQL does with its EXISTS functionality; the rules regarding identifiers are identical, and the inner query also takes a subquery as input.
+This is also very similar in syntax to what SPARQL does with its EXISTS functionality; the rules regarding variables are identical, and the inner query also takes a subquery as input.
 
 == Benefits to this proposal
 
 The existing pattern predicate functionality is very useful, but does not cover all cases.
-Pattern predicates do not allow for introducing identifiers, which makes some queries - such as the one below - difficult to express succinctly:
+Pattern predicates do not allow for introducing variables, which makes some queries - such as the one below - difficult to express succinctly:
 
 [source, cypher]
 ----
--- a/cip/CIP2015-05-13-EXISTS.adoc
+++ b/cip/CIP2015-05-13-EXISTS.adoc
@@ -1,0 +1,162 @@
+= CIP2015-05-13 - EXISTENTIAL SUBQUERIES
+:numbered:
+:toc:
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Authors:* Andrés Taylor <andres@neotechnology.com>
+
+
+[abstract]
+.Abstract
+--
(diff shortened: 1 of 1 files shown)


systay:
Comments shared during the CLG meeting:

 * Change `EXISTS` for properties to be equivalent to `IS NOT NULL` - i.e. valid for any expression and only returning true for null values, and not particular and special for graph elements.
* All uses of `has()` should be capitalized
* Comment on the fact that UNION subqueries will contain RETURN clauses
* Weird layout around line 22
* Add example of valid UNION subquery
* Add example of valid EXISTS use for checking null and not property
* Comment that trailing WITH is only valid together with WHERE

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+To make this feature more powerful, add a new function named +EXISTS+, that would exist in two forms:
+
+ * Property existence checking
+ * Subquery existence checking
+
+=== Syntax
+
+----
+expression = …
```
petraselmer: Better to not use ellipses here (from a readability & comprehension point of view)
May be better to put something like `expression = current definition of expression | bla bla` or something similar (such as in the STARTS WITH CIP)

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+For each matching subgraph evaluated with +EXISTS{...}+, the result value must be true if the subquery finds at least one matching row.
+If no matches are found, false should be returned.
+
+The subquery used in +EXISTS{...}+ has to follow the following rules:
+
+ * It has to be a read-only query. Updating the graph as part of a predicate is not allowed.
+ * It cannot end in a +RETURN+ clause, nor may it contain +UNION+
+ * If +WITH+ ends the subquery, it should be able to change the cardinality of the subquery in an for +EXISTS(...)+ interesting way.
+In other words, they should include +WHERE+ or +SKIP+, since they could turn a query that returns rows into one that does not.
```
petraselmer: It is not easy to understand what "they" means here. Please make clearer, so that the reader feels anchored.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+==== Simple subquery exists
+When the subquery can be described with a single +MATCH-WHERE+ clause, the +MATCH+ keyword can be omitted.
+The difference between this form and a simple pattern predicate, which is already available is that this form allows for introducing new variables inside the +EXISTS(...)+ scope.
```
petraselmer: "...between this form and a simple pattern predicate, which is already available is that this form" -> "between this form and a simple pattern predicate, which is already available*,* is that this form" (addition of the comma)

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+=== Examples
+_Example 1A:_
+
+Return all nodes that have a property named slogan.
```
petraselmer: Put slogan in quotes (easier to read)

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+All forms of +EXIST(...)+ accomplish the same task - check if a particular pattern exists in the database.
+They are expressions, and as such must be side-effect free; that is, the subqueries in +EXIST+ must not be updating queries.
+Regarding nomenclature, all forms of +EXISTS+ are scalar expressions - they return a single boolean value, and this value is never +null+.
+
+==== Property exists
+
+In the property checking form, any expression is accepted as the input.
+If the input expression returns a +null+ value, +EXISTS(...)+ returns false, and for all other values, it will return +true+.
+This form of +EXISTS(...)+ is equivalent to checking for +null+ using +IS NOT NULL+; i.e. +EXISTS(expr)+ is equivalent to +expr IS NOT NULL+.
```
thobe: Nit picking: I would phrase this slightly differently.

    If the input expression would have produced a +null+ value, +EXISTS(...)+ will produce +false+, for all other values it will produce +true+.

1. it doesn't actually return (unless the expression occurs in the `RETURN` clause)
2. I prefer to say _would produce_ rather than _produces_, since the actual implementation might execute the `EXISTS` in a different way than actually evaluate the input expression and checking it for `null`.

thobe: The notion that _any expression_ can be used as a "parameter" for `EXISTS` makes me think about silly expressions such as `EXISTS(EXISTS(foo.bar))`

cip/CIP2015-05-13-EXISTS.adoc
thobe: I agree that something needs to change in this section. It took me several reads before I understood it.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,174 @@
...
+RETURN actor
+----
+
+_Example 1B:_
+Return all nodes with both first name and second name:
+[source, cypher]
+----
+MATCH (actor)
+WHERE EXISTS(actor.firstName + actor.lastName)
+RETURN actor
```
thobe: Wow, this is an interesting side effect of the semantics of `null`.

But in this case I guess it would only work if the two properties are of compatible types. In this example they are likely both `string?`, but what if one was `string?` and the other `number?`?

Oh wait... Cypher does disgusting type coercion to string when one of the operands to `+` is a string... :/

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,174 @@
...
+Find all actors who have acted together with another actor with the same name.
+
+[source, cypher]
+----
+MATCH (actor:Actor)
+WHERE EXISTS {
+	MATCH (other:Actor)-[:ACTED_IN]->(movie)<-[:ACTED_IN]-(actor)
+	WHERE other.name = actor.name
+}
+RETURN actor
```
thobe: Could we make this example a different one from the previous. Such that this example _requires_ the `MATCH`?

cip/CIP2015-05-13-EXISTS.adoc
systay: You are right. This is not a good example.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Authors:* Andrés Taylor <andres@neotechnology.com>
+
+
+[abstract]
+.Abstract
+--
+This CIP introduces EXISTS, a general function that can test for the existence of properties, simple patterns and full subqueries.
```
petraselmer: Can `EXISTS` be backticked, so that we gave formatting? Will make it nicer on the eye...

petraselmer: Can the distinction between when `exists` is a function and when `EXISTS` is a keyword be made explicit somewhere?

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+            ;
+
+property exists = "exists", "(", expression ")"
+                ;
+
+subquery exists = "EXISTS", "{", read only clause, { read only clause }, "}" ;
+
+simple subquery exists = "EXISTS", "{", simple match, "}" ;
+
+simple match = pattern, { ",", pattern }, [ "WHERE", predicate ] ;
```
Mats-SX: Using spaces in these rule names trip me up; can we use camel case please?

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+.Abstract
+--
+This CIP introduces `EXISTS`, a function and a keyword for checking the existence of properties, simple patterns and full subqueries.
+--
+
+toc::[]
+
+== Background
+
+In a way, Cypher already has existential subqueries - pattern predicates allow checking for the existence of subgraphs.
```
petraselmer: predicates allow checking --> predicates checking

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+This form is helpful, but not exhaustive in the types of subqueries that users want to be able to express.
+
+== Proposal
+
+To make this feature more powerful, add a new function named `exists`, and a keyword named the same, which allows for two different predicates:
```
Mats-SX: This sentence is a bit weird. I suggest 

To make this feature more powerful, this CIP suggests the addition of a new function `exists()`, and a keyword `EXISTS`, allowing for two different predicates:

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+expression = <current definition of expression>
+           | property exists
+           | subquery exists
+           | simple subquery exists
+            ;
+
+property exists = "exists", "(", expression ")"
+                ;
+
+subquery exists = "EXISTS", "{", read only clause, { read only clause }, "}" ;
```
Mats-SX: `readOnlyClause` is not defined. At the very least, it should have a prose definition (e.g. in a comment). At the best, an exhaustive list of clauses.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+simple subquery exists = "EXISTS", "{", simple match, "}" ;
+
+simple match = pattern, { ",", pattern }, [ "WHERE", predicate ] ;
+
+----
+Note: the read-only clause above includes WITH but not RETURN.
```
Mats-SX: This note would be obsoleted by a better definition of `readOnlyClause` as suggested above.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+=== Semantics
+
+All forms of `EXIST(...)` accomplish the same task - check if a particular pattern exists in the database.
```
Mats-SX: I suggest changing this to 

Both forms of `EXISTS` accomplish the same task: checking whether a particular pattern exists in the graph.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+They are expressions, and as such must be side-effect free; that is, the subqueries in `EXISTS` must not be updating queries.
+All forms of `EXISTS` are scalar expressions - they return a single boolean value, and this value is never `null`.
+
+==== Property exists
+
+In the property checking form, any expression is accepted as the input.
+If the input expression would have produced a `null` value, `exists(...)`` evaluates to false, and for all other values, it will return `true`.
```
Mats-SX: I suggest changing this to

If the input expression produces `null`, `exists()` will evaluate to `false`. For any other value, it will evaluate to `true`.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+This form of `exists(...)` is equivalent to checking for `null` using `IS NOT NULL`; i.e. `exists(expr)` is equivalent to `expr IS NOT NULL`.
```
Mats-SX: Here, and in the rest of the document, the function-version of the proposed feature should consistently use `exists()` as reference (supported by the style guide).

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+The difference between this form and a simple pattern predicate, which is already available, is that this form allows for introducing new variables inside the `EXISTS(...)` scope.
+
+=== Examples
+_Example 1A:_
+
+Return all nodes that have a property named `slogan`.
+[source, cypher]
+----
+MATCH (actor)
+WHERE EXISTS(actor.slogan)
```
Mats-SX: Functions are written in camelCase: `exists(action.slogan)`

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+They are expressions, and as such must be side-effect free; that is, the subqueries in `EXISTS` must not be updating queries.
+All forms of `EXISTS` are scalar expressions - they return a single boolean value, and this value is never `null`.
+
+==== Property exists
+
+In the property checking form, any expression is accepted as the input.
+If the input expression would have produced a `null` value, `exists(...)`` evaluates to false, and for all other values, it will return `true`.
+This form of `exists(...)` is equivalent to checking for `null` using `IS NOT NULL`; i.e. `exists(expr)` is equivalent to `expr IS NOT NULL`.
+
+This form is there to make property existence checking idiomatic in Cypher - we do not want the user thinking of missing properties as checking for `null` values.
```
Mats-SX: I do not understand this sentence. We do not want the user thinking of 'missing properties' as what exactly? 'Checking for null values' doesn't seem to be a noun.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+ * Subquery existence checking
+
+=== Syntax
+
+----
+expression = <current definition of expression>
+           | property exists
+           | subquery exists
+           | simple subquery exists
+            ;
```
Mats-SX: Misaligned semicolon.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+For each matching subgraph evaluated with `EXISTS{...}`, the result value must be true if the subquery finds at least one matching row.
+If no matches are found, false should be returned.
+
+The subquery used in `EXISTS{...}` has to follow the following rules:
+
+ * It has to be a read-only query. Updating the graph as part of a predicate is not allowed.
+ * It cannot end in a `RETURN` clause, nor may it contain `UNION`
+ * If `WITH` ends the subquery, it should be able to change the cardinality of the subquery.
+ In other words, the `WITH` clause should include `WHERE` or `SKIP`, since these sub-clauses can turn a query that returns rows into one that does not.
```
Mats-SX: I'd say that the `WITH` clause _could_ include `WHERE` or `SKIP`, rather than _should_, right?

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+Find all actors who have acted together with another actor with the same name on at least two movies.
+
+[source, cypher]
+----
+MATCH (actor:Actor)
+WHERE EXISTS {
+	MATCH (other:Actor)-[:ACTED_IN]->(movie)<-[:ACTED_IN]-(actor)
+	WHERE other.name = actor.name
+  WITH other, count(*) as c
+  WHERE c > 1
```
Mats-SX: Weird indentation.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+}
+RETURN actor
+----
+
+=== Interaction with existing features
+This replaces the pre-existing `HAS(...)` function.
```
Mats-SX: `has()` is already gone, so I suggest removing this part completely. Instead, it could be mentioned that pattern predicates will be obsoleted by this feature, and that they subsequently should be deprecated/removed.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+== What others do
+
+This is very similar to what SQL does with its EXISTS functionality.
```
Mats-SX: backticks around `EXISTS`

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+
+This is also very similar in syntax to what SPARQL does with its EXISTS functionality; the rules regarding variables are identical, and the inner query also takes a subquery as input.
```
Mats-SX: Backticks for `EXISTS`

Mats-SX:
@systay I have made a commit with amendments according to these comments (after discussing this CIP with the CLG), but I can not push it to your fork. Please add me as a collaborator.

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,167 @@
...
+==== Property exists
+
+In the property checking form, any expression is accepted as the input.
+If the input expression would have produced a `null` value, `exists(...)`` evaluates to false, and for all other values, it will return `true`.
+This form of `exists(...)` is equivalent to checking for `null` using `IS NOT NULL`; i.e. `exists(expr)` is equivalent to `expr IS NOT NULL`.
+
+This form is there to make property existence checking idiomatic in Cypher - we do not want the user thinking of missing properties as checking for `null` values.
+
+==== Subquery exists
+When checking for the existence of subqueries, variables and parameters from the outer query are available in the subquery.
```
petraselmer: Perhaps say something about this being a correlated subquery? We've done the same for the 'general' subqueries CIPed elsewhere...

cip/CIP2015-05-13-EXISTS.adoc
petraselmer: Is there a reason why `EXISTS {...}` cannot be nested? At the very least, this ought to be mentioned.

Mats-SX:
Replaced by #120.

PR #6, opened 2015-10-13
Added an updated version of the state visibility CIP

systay:
Some github weirdness is making rendering the AsciiDocs problematic. Here is a link to the interesting bits:

https://github.com/systay/cypher/blob/c8d576cb2b4f08fd43b9cf43b15dc539f773c507/cip/CIP2015-10-03-State-visibility-between-clauses.adoc

PR #11, opened 2015-10-27
Adds state visibility CIP

systay:
Replaces #6

### DIFF
--- a/cip/CIP2015-10-03-State-visibility-between-clauses.adoc
+++ b/cip/CIP2015-10-03-State-visibility-between-clauses.adoc
@@ -1,0 +1,76 @@
+= CIP20140109 - State visibility between clauses
+:numbered:
+:toc:
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Authors:* Andrés Taylor <andres@neotechnology.com>
+
+[abstract]
+.Abstract
+--
+This CIP tries to formalise how state changes behave while inside a Cypher query. Currently, because this is not explicit enough, the behaviour is implementation dependent.
+--
+
+toc::[]
+
+=== Background
+Cypher allows queries to go through multiple steps, and it’s possible to interleave reading and writing to the graph.
+Some clauses can also do both - read the graph and write to it at the same time.
+
+To allow the user to understand what the outcome will be at an abstract, logical level, semantics for the visibility of these changes should be clear.
+
+=== Proposal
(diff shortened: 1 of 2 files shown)


cip/CIP2015-10-03-State-visibility-between-clauses.adoc
```diff
@@ -0,0 +1,76 @@
...
+To allow the user to understand what the outcome will be at an abstract, logical level, semantics for the visibility of these changes should be clear.
+
+=== Proposal
+Each clause lives in its own state, which includes all the changes of clauses coming before it, including changes performed by itself, *but none of the changes made by clauses listed later in the query.*
+
+The semantical data flow here is such that each clause operates on the entire result set before passing control to any subsequent clause.
+
+Updates between UNION subqueries are also treated in a sequential manner.
+Therefore subqueries coming later will see updates from subqueries before, but not the other way around.
+
```
petraselmer: I think all keywords ought to be formatted, so please change "UNION" -> "`UNION`"

cip/CIP2015-10-03-State-visibility-between-clauses.adoc
```diff
@@ -0,0 +1,76 @@
...
+==== Examples
+
+.Example query 1:
+The following query would lead to a never-ending loop, if a lazy +MATCH+ clause is not isolated from the new nodes created by +CREATE+.
```
petraselmer: It seems that "+bla+" doesn't work, as there is no formatting. Rather use backticks.

systay:
Comments shared during CLG meeting:

* Point out explicitly that this is not about concurrency
* Add example with UNION
* Rename this CIP instead of reusing the CIP id

cip/CIP2015-10-27-State-visibility-between-clauses.adoc
```diff
@@ -0,0 +1,94 @@
...
+toc::[]
+
+=== Background
+Cypher allows queries to go through multiple steps, and it’s possible to interleave reading and writing to the graph.
+Some clauses can also do both - read the graph and write to it at the same time.
+
+To allow the user to understand what the outcome will be at an abstract, logical level, semantics for the visibility of these changes should be clear.
+
+Note:
+This is not about concurrency - even in a single user system this would need to be thought through and defined explicitly._
```
thobe: remove trailing underscore

PR #4, opened 2015-10-12
Public Type System and Type Annotations [OLD]

boggle:

PR #12, opened 2015-11-09
Public Type System and Type Annotations

boggle:


boggle:
Original PR & Discussion: https://github.com/opencypher/openCypher/pull/4

cip/CIP2015-09-16-public-type-system-type-annotation.adoc
```diff
@@ -0,0 +1,407 @@
...
+
+The details of internal types are out of scope for the purposes of this document.
+
+== Proposal
+
+=== Types and Type Literal Syntax
+
+The section below presents presents the new Public Type System (PTS) proposed for Cypher.
+
+==== Material and nullable types
```
cleishm: It would be helpful to discuss _why there is a need for (non)nullable types_. Requiring non-null values is typically a runtime concern in a language. The current Neo4j implementation of cypher treats all types as nullable, and warning/error messages containing type names do not distinguish between them. If this CIP was to be accepted, an error message containing "String" would more correctly, yet oddly, contain "String?".

systay: For the nullable types, would it make sense to add something about how current Cypher constructs relate to nullability? As an example: `MATCH` will never introduce a nullable variable, but `OPTIONAL MATCH` will only introduce nullable variables. This discussion probably should not block the CIP, but it would be interesting to include it.

cip/CIP2015-09-16-public-type-system-type-annotation.adoc
```diff
@@ -0,0 +1,409 @@
...
+* `PATH`
+** A path from a node `n1` to a node `ni` - corresponding to a walk in the graph from `n1` to `ni` - is a sequence `n1`, `r1`, `n2`, `r2`, ..., `r(i-1)`, `ni` of alternating nodes and relationships such that for 1 <= `j` < `i`, any `rj` contained in the path is incident with `nj` and `n(j+1)`. Additionally, a single node path is a path that starts and ends at the same node `n0` and does not contain any relationships.
+
+==== Void type
+
+This CIP also introduces an additional nullable type called `VOID`. `VOID` is intended to be used as the return type of user defined procedures that execute without producing a result value, i.e. the return type of procedures with side effects. `VOID` is a subtype of `ANY?`. In an expression context, `NULL` is the only value that inhibits the `VOID` type, indicating a missing "real" return value (Note though that Cypher currently does not have any expressions of type `VOID` and this is merely defined here for completeness).
+
+==== Reserved type names
```
thobe: Given that this list (or any list like this) is not (going to be) exhaustive. Does it really make sense to include it at all?

cip/CIP2015-09-16-public-type-system-type-annotation.adoc
```diff
@@ -0,0 +1,409 @@
...
+== Caveats to this Proposal
+
+Adding explicit syntax for types may lead to a more complex and difficult to learn language.
+Since the initial use of type syntax is for specifying the signatures of user defined procedures only, this should not be a strong concern.
+
+== Audience Consultation
+
+Communication with external stakeholders should happen as part of the consultation process for _CIP2015-06-24 Managing Procedures_.
+
+== References
```
thobe: Can we use links here?

cip/CIP2015-09-16-public-type-system-type-annotation.adoc
```diff
@@ -0,0 +1,409 @@
...
+
+==== Union type support
+Cypher's current type checker tracks sets of possible types for expressions.
+This could be understood as a form of union typing.
+Union types do not map easily to the type systems of most common target languages (Java, Javascript, ...), they complicate type checking, and are not required for type mapping to common external languages.
+This is why they have not been included in this proposal.
+They could be added instead at a later stage.
+
+*Example*
+`RETURN [1, "Yo"] :: LIST OF STRING | LIST OF NUMBER`
```
thobe: that seems wrong, the example indicates that this is a `list of strings` or a `list of numbers`, when in fact it is a `list of (strings or numbers)`. perhaps it should rather be `RETURN [1, "Yo"] :: LIST OF (STRING | NUMBER)`

cip/CIP2015-09-16-public-type-system-type-annotation.adoc
```diff
@@ -0,0 +1,409 @@
...
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Author:* Stefan Plantikow <stefan.plantikow@neotechnology.com>
+*Author:* Nigel Small <nigel.small@neotechnology.com>
+
+[abstract]
+.Abstract
+
+This CIP specifies the Public Type System (PTS) of Cypher and introduces literal syntax for writing down types as well as new syntax for type annotations. The proposed type annotation syntax does not change the Cypher grammar right now but rather is intended as a template to be used by future CIPs. This CIP also informatively discusses future syntax that might use the introduced literal type syntax.
```
thobe: Why do we need to have an abbreviation for the public type system? I feel that this is unnecessary, and that "PTS" in particular is not a very good abbreviation, since it makes me think of "Post Traumatic Stress" before I think of "Public Type System"...

cip/CIP2015-09-16-public-type-system-type-annotation.adoc
cleishm: Right. I actually think this whole section is wrong - in the type checker it's not a 'union of types', it's a 'set of possible types'. And the example is thus wrong: according to the type checker this example is definitely `LIST OF ANY`. A better example would be:
```
UNWIND [ [1,2,3], ["a", "b", "c"] ] AS x
RETURN x :: LIST OF STRING | LIST OF NUMBER
```

cleishm: Note: we have discussed a type checker that is able to track multiple specific types used within a list. It would probably still give lists a general contained type, but then tag specific offsets as having a more specific typed value. However nothing like that has been implemented.

boggle: Yes, I'll fix that, it's a glitch.

PR #13, opened 2015-11-09
First draft of Type Coercion CIP

Mats-SX:


cleishm:
It might be nice if this document summarised the existing coercions (i.e. what the Neo4j implementation already does) and then also listed a set of alternatives for modifying that, where each alternative may add and/or remove coercions from the specification.

Mats-SX:
@cleishm It does summarise the existing coercions. It's not intended to modify anything, but to clearly document the current language behaviour, as a resource to refer to when considering implementation.

cleishm:
@Mats-SX Ah, right. It's under "proposal".

It might be worth using a different format for this, since there is no proposal - it's capturing what the existing language semantic is. That'll make it more obvious too, since I totally missed that one line (which is actually the point of the whole document!).

Mats-SX:
I'll leave that to the CLG to decide. This CIP is done at their request.

cip/CIP2015-11-09-type-coercions.adoc
```diff
@@ -0,0 +1,98 @@
...
+The current Cypher behavior is proposed, which consists of two type coercions: `CTCollection` to `CTBoolean`, and `CTInteger` to `CTFloat`.
+
+=== Examples
+
+Here follows a few queries in which types are coerced.
+
+Check if a collection is empty, and if it is not, return the first element:
+[source, cypher]
+----
+WITH [] as coll
```
boggle: Could we try to get a handle on how frequent this is used in the wild? My opinion on coercion always has been that we should only coerce if no information is lost and use a cast or a function call otherwise.

cip/CIP2015-11-09-type-coercions.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+No syntax changes are proposed.
+
+=== Semantics
+
+There are no new semantics to consider.
+
+=== Alternatives
+
+There are plenty of additional coercions to consider.
```
boggle: We could limit int to float coercion for literals only; that perhaps is less surprising and addressed the 80/20 of why that coercion is practical.

cip/CIP2015-11-09-type-coercions.adoc
Mats-SX: What information is lost when coercing a collection into a boolean?

cip/CIP2015-11-09-type-coercions.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+In the case of a collection with both integer and floating-point numbers, it's type will be a collection of numbers.
+Functions that work on both integers and floats would then accept all elements of this collection one-by-one, but if they do not also accept type number they will not be able to act on the collection itself when used in a comprehension.
+In this type of situation, it is very valuable for a user to know what kinds of coercions that Cypher will perform, and what explicit casting that they will have to do themselves.
+
+== Proposal
+
+The current Cypher behavior is proposed, which consists of two type coercions: `CTCollection` to `CTBoolean`, and `CTInteger` to `CTFloat`.
+
+=== Examples
```
Mats-SX: We should probably add an example showing how values of lists are coerced:

`WITH [1, 1.0] as list CREATE ({l:list})` will coerce the `LIST OF NUMBER` to `LIST OF FLOAT`, so it then can store it as a `float[]`.

cip/CIP2015-11-09-type-coercions.adoc
boggle: The elements?

cip/CIP2015-11-09-type-coercions.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+== Background
+
+In the case of a collection with both integer and floating-point numbers, it's type will be a collection of numbers.
+Functions that work on both integers and floats would then accept all elements of this collection one-by-one, but if they do not also accept type number they will not be able to act on the collection itself when used in a comprehension.
+In this type of situation, it is very valuable for a user to know what kinds of coercions that Cypher will perform, and what explicit casting that they will have to do themselves.
+
+== Proposal
+
+The current Cypher behavior is proposed, which consists of two type coercions: `CTCollection` to `CTBoolean`, and `CTInteger` to `CTFloat`.
```
boggle: Information is also potentially lost here if the integer doesn't fit the float.

cip/CIP2015-11-09-type-coercions.adoc
Mats-SX: I don't think I understand what you mean when you say that information is lost. The list will still be around, as will its elements. I'm probably missing some piece of education on this subject.

boggle: When you coerce a value A to a value B what I mean by "information is lost/preserved" is basically "can you answer the same questions by just looking at the value only and get the same answers", or perhaps more simple, can you go from A to B and back and arrive at the same A?

Mats-SX: That makes perfect sense. Thank you.

PR #14, opened 2015-11-09
Adds the STARTS WITH and ENDS WITH CIP

petraselmer:
This CIP has already been accepted.

### DIFF
--- a/cip/CIP2015-09-15-STARTS-WITH-and-ENDS-WITH.adoc
+++ b/cip/CIP2015-09-15-STARTS-WITH-and-ENDS-WITH.adoc
@@ -1,0 +1,144 @@
+= CIP2015-09-15 - STARTS WITH and ENDS WITH
+:numbered:
+:toc:
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Author:* Stefan Plantikow stefan.plantikow@neotechnology.com
+
+[abstract]
+.Abstract
+--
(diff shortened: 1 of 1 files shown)


cip/CIP2015-09-15-STARTS-WITH-and-ENDS-WITH.adoc
```diff
@@ -0,0 +1,144 @@
...
+
+ * More complex string searches must still use the regular expression search.
+ * Differs from SQL's approach.
+ * More keywords added to the language.
+
+== Appendix
+
+=== Background on case-insensitive search
+
+Case-insensitive string search requires specifying a case folding function for converting the string operands as well as using a suitable equality predicate to compare them. Unicode itself defines three possible cases: lower case, upper case, and title case. Converting to these cases is inherently locale specific though in practice this is often ignored by using the default ("C" or "en") locale. To achieve good results, this is often combined with another locale dependent normalization step. Furthermore notably Java defines a special equality predicate "equalsIgnoreCase" for case insensitive comparison that treats two strings identical if they have the same length and if all of their characters are pairwise equal either directly, or after upcasing both of them, or after downcasing both of them.
```
boggle: We should replace `case folding function` with `case conversion function` as case folding has a special meaning (which I didn't knew when writing this). Also replace `locale dependent normalization step` with `locale independent normalization step`. Makes more sense this way.

petraselmer: Rectified...

PR #15, opened 2015-12-10
CIP2015-08-06 Date and Time

thobe:
[CIP2015-08-06](https://github.com/thobe/openCypher/blob/date-time/cip/1.accepted/CIP2015-08-06-date-time.adoc)

### DIFF
--- a/cip/1.accepted/CIP2015-08-06-date-time.adoc
+++ b/cip/1.accepted/CIP2015-08-06-date-time.adoc
@@ -1,0 +1,985 @@
+= CIP2015-08-06 - Date and Time
+:numbered:
+:toc:
+:toc-placement: macro
+:source-highlighter: codemirror
+ifdef::env-github,env-browser[:outfilesuffix: .adoc]
+
+*Author:* Tobias{nbsp}Lindaaker{nbsp}<tobias@neotechnology.com>
+
+[abstract]
+.Abstract
+--
+We propose to introduce new types for dealing with temporal data.
+The main part of this document, the <<proposal,Proposal section>>, is written in present tense, as if Cypher already supported the proposed features.
+If this document appears too lengthy to you, dear reader, please refer to the <<impl-notes,implementation notes section>> for a motivation of why this proposal despite all its length and detail would not be troublesome to implement.
+--
+
+toc::[]
+
+
+== Background
+
+Dealing with dates and times is a long standing feature request for Cypher.
+With Cypher having become the main means of interacting with a graph database, it becomes crucial to have built in support for handling date and time in order to support queries with such needs.
+
+Most other database management systems and query languages, relational database in particular, support dealing with temporal data.
+The addition of this support to Cypher would bridge that feature gap.
+
+Since Cypher is the main way of interacting with the database, this document takes a Cypher-centric approach to dates and time.
+These additions do however have implications to other parts of Neo4j as well.
+Where appropriate we will make references to other documents describing those concerns.
+
+[[proposal]]
+== Proposal
+
+Cypher has built in support for handling temporal values, and the underlying database supports storing these temporal values as properties on nodes and relationships.
+
+
+=== Temporal types
+
+Cypher supports 6 different temporal value types, in two different categories.
+The first category contains the _temporal instant_ types: _DateTime_, _LocalDateTime_, _Date_, _Time_, and _LocalTime_.
+The second category contains a single type, _Duration_.
+The _temporal instant_ values all express a point in time, an instant, with varying degrees of precision, and a _Duration_ represents an amount of time.
+
+
+==== DateTime
+
+An instant capturing the date, the time, and the timezone.
+
+
+==== LocalDateTime
+
+An instant capturing the date and the time, but not the timezone.
+
+
+==== Date
+
+An instant capturing the date, but not the time, nor the timezone.
+
+
+==== Time
+
+An instant capturing the time of day, and the timezone, but not the date.
+
+
+==== LocalTime
+
+An instant capturing the time of day, but not the date, nor the timezone.
+
+
+==== Duration
+
+A temporal amount.
+This captures the difference in time between two instants.
+It only captures the amount of time between two instants, it does not capture a start time and end time.
+A unit capturing the start time and end time would be a _Time Interval_ and is out of scope for this proposal. +
+A _Duration_ can be negative.
+
+
+=== Temporal functions
+
+For producing values of each temporal type, Cypher has a function corresponding to each of the types.
+These functions bear the same name as the type.
+These functions construct the type they correspond to in one of four ways, either through composition of the components of the type, or through parsing of a string representation of the temporal value, or through selection and composition of components from another temporal value, or by returning the current time.
+
+In order to act in these different ways there are three different types of parameters that can be passed to the temporal functions:
+
+ * By passing no parameters, the current time is returned.
+   Where _“current time”_ is defined to be the same throughout all invocations within the same query.
+ * When passing a single string parameter, the function <<parsing,parses a temporal value>> from the string.
+ * Finally _named parameters_ can be passed to the functions by passing a single map parameter.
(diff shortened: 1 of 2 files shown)


boggle:
Could you please ensure that this aligns with the type system and the equality CIPs and update them if needed as part of this PR?

cip/1.accepted/CIP2015-08-06-date-time.adoc
```diff
@@ -379,8 +394,15 @@ Furthermore these components are available for specifying time:
...
-* `datetime.fromepochmillis($millis)`
+A common pattern in many programming languages and data sets is to represent time as a number representing the amount of time that has elapsed from the UNIX epoch footnote:[UNIX epoch: `1970-01-01T00:00:00` (UTC)].
+In order to work with such data, Cypher provides functions for converting such a numeric timestamp to a _DateTime_ value:
+
+* `datetime.fromepoch($seconds,$nanos)` - Creates a _DateTime_ value at the specified number of _seconds_ and _nanoseconds_ from the UNIX epoch in the UTC timezone.
+* `datetime.fromepochmillis($millis)` - Creates a _DateTime_ value at the specified number of _milliseconds_ from the UNIX epoch in the UTC timezone.
+
+Conversions to other temporal types from UNIX epoch representation can be achieved by going via _DateTime_.
+
+Cypher also provides <<instant-components,accessors>> for converting _to_ UNIX epoch representation.
```
sherfert: This is true only for datetime.

cip/1.accepted/CIP2015-08-06-date-time.adoc
```diff
@@ -542,42 +567,56 @@ Select the temporal instant corresponding to the millenium of the given instant.
...
+3+| *Example:* `date.truncate('decade', date("1986-04-26"))` _yields_ `date("1980-01-01")`
 
 .3+| `year` +
 Select the temporal instant corresponding to the year of the given instant.
 | `datetime.truncate('*year*', _input_)` | _DateTime_, _LocalDateTime_^†^, _Date_^†^
 | `localdatetime.truncate('*year*', _input_)`, | _DateTime_, _LocalDateTime_, _Date_
 | `date.truncate('*year*', _input_)` | _DateTime_, _LocalDateTime_, _Date_
+3+| *Example:* `date.truncate('year', date("2015-08-21"))` _yields_ `date("2015-01-01")`
+
+.3+| `weekYear` +
```
sherfert: This is not covered in the implementation so far.

cip/1.accepted/CIP2015-08-06-date-time.adoc
```diff
@@ -785,42 +835,49 @@ Components of temporal values are accessed as properties.
...
+|`second`        | yes | yes | no  | yes | yes
+|`millisecond`   | yes | yes | no  | yes | yes
+|`microsecond`   | yes | yes | no  | yes | yes
+|`nanosecond`    | yes | yes | no  | yes | yes
+|`timezone`      | yes | no  | no  | yes | no
+|`offset`        | yes | no  | no  | yes | no
+|`offsetMinutes` | yes | no  | no  | yes | no
+|`offsetSeconds` | yes | no  | no  | yes | no
+|`epochMillis`   | yes | no  | no  | no  | no
+|`epochSeconds`  | yes | no  | no  | no  | no
```
sherfert: The previous four accessors are different from the current implementation.

cip/1.accepted/CIP2015-08-06-date-time.adoc
```diff
@@ -915,11 +972,11 @@ The temporal types form a separate type system within the cypher type system, th
 
 The previous timestamp function of Cypher still works, and returns the equivalent value of
 
-    datetime().epoch
+    datetime().epochMillis
```
sherfert: Renamed accessor.

cip/1.accepted/CIP2015-08-06-date-time.adoc
```diff
@@ -915,11 +972,11 @@ The temporal types form a separate type system within the cypher type system, th
...
 
 Perhaps more importantly a value produced by the old +timestamp+ function can be converted to a _DateTime_ by passing it as the `epoch` component to the `datetime` function:
 
-    datetime({epoch:timestamp()})
+    datetime.fromEpochMillis(timestamp())
```
sherfert: Implementation note: We can still support the old way (as a value in a map), but we should also support this new way!

Tobias note: The sentence above does not reflect this changes.

cip/1.accepted/CIP2015-08-06-date-time.adoc
```diff
@@ -785,42 +835,49 @@ Components of temporal values are accessed as properties.
...
+|`hour`          | yes | yes | no  | yes | yes
+|`minute`        | yes | yes | no  | yes | yes
+|`second`        | yes | yes | no  | yes | yes
+|`millisecond`   | yes | yes | no  | yes | yes
+|`microsecond`   | yes | yes | no  | yes | yes
+|`nanosecond`    | yes | yes | no  | yes | yes
+|`timezone`      | yes | no  | no  | yes | no
+|`offset`        | yes | no  | no  | yes | no
+|`offsetMinutes` | yes | no  | no  | yes | no
+|`offsetSeconds` | yes | no  | no  | yes | no
```
sherfert: The property store does not even support seconds. Neither do drivers. Therefore, I think it does not make sense to have this accessor for Neo4j. Should we keep it in this proposal and document in Neo4j that in this respect we diverged from openCypher?

cip/1.accepted/CIP2015-08-06-date-time.adoc
thobe: I'm don't think we should support the value in a map way for this.

thobe: After thinking this through, we should go back to the map-based form, but rename the component to `epochMillis`, since this allows specifying a timezone.

cip/1.accepted/CIP2015-08-06-date-time.adoc
```diff
@@ -785,6 +785,12 @@ Does lhs occur after rhs:
     WITH date(...) AS lhs, date(...) AS rhs // or another temporal instant type
     RETURN lhs > rhs
 
+Instant values that occur at the same point in time, but that have a different timezone are not considered equal, and must therefore be ordered in some predictable way.
+Cypher prescribes after the primary order of point in time, instant values be ordered by effective timezone offset, from west (negative offset from UTC) to east (positive offset from UTC).
+This has the effect that times that represent the same point in time will be ordered with the time with the earliest local time first.
+If two instant values represent the same point in time, and have the same timezone offset, but a different named timezone (this is possible for _DateTime_ only, since _Time_ only has an offset), these values are not considered equal, and ordered by the timezone identifier, alphabetically, as its third ordering component.
```
sherfert: If a has just an offset (no named timezone) and b has a named timezone (and everything else applies, i.e. same instant in time and same effective offset), then a < b.

cip/1.accepted/CIP2015-08-06-date-time.adoc
```diff
@@ -785,6 +785,12 @@ Does lhs occur after rhs:
     WITH date(...) AS lhs, date(...) AS rhs // or another temporal instant type
     RETURN lhs > rhs
 
+Instant values that occur at the same point in time, but that have a different timezone are not considered equal, and must therefore be ordered in some predictable way.
+Cypher prescribes after the primary order of point in time, instant values be ordered by effective timezone offset, from west (negative offset from UTC) to east (positive offset from UTC).
+This has the effect that times that represent the same point in time will be ordered with the time with the earliest local time first.
```
sherfert: This is not true for offsets greater than +12:00 or smaller than -12:00 (only true for Time, not for DateTime).

thobe: True. It could be clarified that it has that effect for _DateTime_.

PR #17, opened 2016-01-27
Add MANDATORY MATCH

boggle:


cip/CIP2016-01-26-mandatory-match.adoc
```diff
@@ -0,0 +1,72 @@
...
+== Background
+
+The idea for this CIP grew out of a (now abandoned) dog fooding project for storing benchmarking information in Neo4j.
+
+== Proposal
+
+This CIP suggests the introduction of a new clause called `MANDATORY MATCH` that has the same basic form as the existing `MATCH` clause (i.e. `MANDATORY MATCH <pattern> [WHERE <predicate>]`).
+
+If matching the graph pattern described by `MANDATORY MATCH` fails to match at least a single row, the execution of the while Cypher query fails with a descriptive error message that points to which `MANDATORY MATCH` clause failed and
+includes all bound identifiers and parameters used in trying to match the described graph pattern.
```
ikwattro: typo : of the `whole` Cypher query

ikwattro:
By reading the proposal, it seems to be based on use cases where unique constraints exists, what is the expected behavior of such clause if no unique constraint exist but well an index and more than one row is found ?

cip/CIP2016-01-26-mandatory-match.adoc
```diff
@@ -0,0 +1,72 @@
...
+.Abstract
+--
+This CIP proposes adding a new variant of the `MATCH` clause called `MANDATORY MATCH` that fails when it does not match at least a single row in the underlying graph.
+--
+
+toc::[]
+
+== Motivation
+
+When writing applications using Cypher, many queries will start off by looking up multiple individual nodes based on an unique id with the implied expectation of matching exactly one node. When that expectation is not fulfilled (either due to sending a wrong id or the node having been deleted) it becomes important to identify which lookup failed. Achieving  this currently is not well supported by Cypher.
```
cleishm: The same can be said of non-unique matches, or of uses of clauses like UNWIND. Has there been any consideration of a general assertion mechanism? There's a lot of similarity between this concept and the concept of SKIP/LIMIT: `LIMIT n` reduces the number of rows to be [0..n]. I could imagine a variation on that which _requires_ the number of rows to be within a specified range (e.g. `REQUIRE 1`, `REQUIRE 1..`, `REQUIRE 1..4`, etc.)

boggle:
@ikwattro That's fine; you're right this is mainly for use with unique constraints or indices. It is intended to just make sure that at least one row is being returned. I've seen quite a few application queries that basically start by doing a lot of index lookups and if any of them returns no rows (because of an invalid argument for example) then there's not really great error reporting regarding *which* match did not return any data. This proposal tries to address this.

boggle:
@cleishm interesting and probably worth keeping in mind for the syntax but the use case really is to ensure >= 1 rows. I have seen no practical need to do anything else.

cip/CIP2016-01-26-mandatory-match.adoc
thobe: @cleishm: I agree, some general `REQUIRE` construct could be useful. But I think this is a different thing from what `MANDATORY MATCH` is trying to do. `MANDATORY MATCH` talks about a specific pattern having to match something, or the query will fail, whereas `REQUIRE` would talk about a required cardinality at a particular point in the query. `MANDATORY MATCH` makes the matching required for _each incoming row_, emulating that with `REQUIRE` would require the user to put the `MATCH ... REQUIRE` in a sub-query, at which point I believe `MANDATORY MATCH` is a more intuitive way of expressing the intent.

cip/CIP2016-01-26-mandatory-match.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+== Background
+
+The idea for this CIP grew out of a (now abandoned) dog fooding project for storing benchmarking information in Neo4j.
```
thobe: I actually feel that these details are unnecessary. What would be relevant is describing what problem this proposal solves (the previous section already does that) and when that problem arises.

cip/CIP2016-01-26-mandatory-match.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+== Proposal
+
+This CIP suggests the introduction of a new clause called `MANDATORY MATCH` that has the same basic form as the existing `MATCH` clause (i.e. `MANDATORY MATCH <pattern> [WHERE <predicate>]`).
+
+If matching the graph pattern described by `MANDATORY MATCH` fails to match at least a single row, the execution of the whole Cypher query fails with a descriptive error message that points to which `MANDATORY MATCH` clause failed and includes all bound identifiers and parameters used in trying to match the described graph pattern.
+
+=== Example
+
+Find some nodes for which uniqueness constraints have been created up-front:
```
thobe: Whether there is a uniqueness constraint or not is irrelevant for `MANDATORY MATCH`, is it not?

cip/CIP2016-01-26-mandatory-match.adoc
```diff
@@ -0,0 +1,74 @@
...
+However in the interest of keeping this proposal small and specific, this is not considered here.
+
+== What others do
+
+No similar facility exists in SQL.
+
+== Benefits to this proposal
+
+Application programmers get a new powerful facility for detecting semantical errors in their applications while
+also failing early in the case of an error (thus avoiding unnecessary work).
```
thobe: I don't think this is only about verifying parameters, surely there must be other reasons/motivations for introducing this concept?

Mats-SX:
Needs to be rebased to end up in the correct directory (`accepted`).

cip/CIP2016-01-26-mandatory-match.adoc
boggle: True, this just refers back to the originally motivating use-case.

cip/CIP2016-01-26-mandatory-match.adoc
boggle: Ok, I'll remove that.

cip/CIP2016-01-26-mandatory-match.adoc
boggle: No, this is the main motivation: Helping a major use-case for typical OLTP queries (given some user-provided primary keys for entities, find data in the graph) by reducing the parameter checking burden on the client side (if even feasible to do there), producing better error messages when wrong parameters have been supplied, and as a side benefit, avoid unnecessary work by failing early (may be more relevant for OLAP style queries).

PR #21, opened 2016-02-29
Initialize the TCK

Mats-SX:
This commit adds a place for the TCK to exist, a readme to describe its
format, and one initial feature.

NOTE: Don't merge this until we've settled for a licensing solution.


petraselmer:
We have - for now - resolved the licensing issue by retaining the usual GPL header. This can be altered in the future, if the need arises.

PR #25, opened 2016-03-17
Add `ExpressionAcceptance` feature

Mats-SX:


tck/README.adoc
```diff
@@ -43,13 +43,23 @@ Scenario: Returning a single node
...
 ----
-<1> This step specifies the required initial state of the graph. `<GRAPH_NAME>` is the name of one of the specified graphs for initial states (read more below).
+[source,gherkin]
+.This scenario illustrates the effects of executing a query `Q` - matching all nodes and returning one of them - on the graph <GRAPH_NAME>.
+----
+Scenario: Indexing a list with a string
+    Given any graph // <1>
+    When executing query: WITH [0] AS expr, 'x' AS idx RETURN expr[idx] // <4>
+    Then a TypeError should be raised at runtime: ListElementAccessByNonInteger // <7>
+----
```
petraselmer: I think that these numbers at the end of each line are very important..but hard to read. Could they be made bold or something so that they stand out a bit more?

tck/README.adoc
```diff
@@ -101,3 +112,28 @@ For this reason, the rows of the expected results table are to be considered a s
...
 
 // TODO: Mention Cucumber ?
+
+[[errors]]
+=== Cypher errors
+
+The `Then` step used to specify expected errors from running a given invalid query follows this schematic setup:
+
+ Then a TYPE should be raised at TIME: DETAIL
+
```
petraselmer: I don't think `TIME` is the best term here - I keep interpreting it as "12:45" and co. So, I suggest a less loaded term... maybe something like `QUERY_STAGE`, `EXECUTION_STAGE`, `ENVIRONMENT` and so on (I am a lot less keen on the latter term than the first two, but have added it in case it inspires something better).

tck/README.adoc
```diff
@@ -101,3 +112,28 @@ For this reason, the rows of the expected results table are to be considered a s
...
+TYPE will be one of the following error types:
+
+// these are initially taken from Neo4j kernel's Status.Statement class' ClientErrors
+- SyntaxError                    "The statement contains invalid or unsupported syntax."
+- SemanticError                  "The statement is syntactically valid, but expresses something that the database cannot do."
+- ParameterMissing               "The statement is referring to a parameter that was not provided in the request."
```
petraselmer: "The statement is referring to a parameter..." -> "The statement refers to a parameter..."

tck/README.adoc
```diff
@@ -101,3 +112,28 @@ For this reason, the rows of the expected results table are to be considered a s
...
+- ConstraintVerificationFailed   "A constraint imposed by the statement is violated by the data in the database."
+- EntityNotFound                 "The statement is directly referring to an entity that does not exist."
```
petraselmer: "The statement is directly referring to an entity that does not exist." -> "The statement refers to a non-existent entity."
Is there any reason to include "directly" in this? I think it is self-evident that it does refer 'directly'...

tck/README.adoc
```diff
@@ -101,3 +112,28 @@ For this reason, the rows of the expected results table are to be considered a s
...
+- PropertyNotFound               "The statement is referring to a property that does not exist."
```
petraselmer: "The statement is referring to a property that does not exist." -> "The statement refers to a non-existent property."

tck/README.adoc
```diff
@@ -101,3 +112,28 @@ For this reason, the rows of the expected results table are to be considered a s
...
+- LabelNotFound                  "The statement is referring to a label that does not exist."
```
petraselmer: "The statement is referring to a label that does not exist." -> "The statement refers to a non-existent label."

tck/README.adoc
```diff
@@ -101,3 +112,28 @@ For this reason, the rows of the expected results table are to be considered a s
...
+- TypeError                      "The statement is attempting to perform operations on values with types that are not supported by the operation."
+- ArgumentError                  "The statement is attempting to perform operations using invalid arguments."
+- ArithmeticError                "Invalid use of arithmetic, such as dividing by zero."
```
petraselmer: "Invalid use of arithmetic, .." -> "Invalid use of an arithmetic operation, .."

tck/README.adoc
Mats-SX: These texts all come from within the kernel, but I'll change them there too. I agree, there's no need for the word 'direct' here. I even think it is slightly misleading; you may get that status code due to corrupted data, without any sort of 'direct' referral from your query.

PR #27, opened 2016-03-18
Add new OrderBy feature file

petraselmer:


tck/features/OrderByAcceptance.feature
```diff
@@ -0,0 +1,151 @@
...
+      | limit | -1 |
+    When executing query: MATCH (p:Person) RETURN p.name AS name ORDER BY p.name LIMIT {limit}
+    Then the result should be:
+      | name |
+
+  //This will only work when https://github.com/opencypher/openCypher/pull/25 is merged...
+  Scenario: ORDER BY with a negative LIMIT should fail with a syntax exception
+    And having executed: CREATE (s:Person {name: 'Steven'}), (c:Person {name: 'Craig'})
+    When executing query: MATCH (p:Person) RETURN p.name AS name ORDER BY p.name LIMIT -1
+    Then a SyntaxError should be raised at compile time: [detail]
```
Mats-SX: That dependency is actually not true. It was really waiting for the implementation of this feature, which is made here: https://github.com/neo4j/neo4j/pull/6735. However, this feature might require the implementation of another `detail` message in `neo4j`, look here: https://github.com/neo4j/neo4j/blob/3.1/community/cypher/compatibility-suite/src/test/scala/cypher/feature/steps/CypherTCKSteps.scala#L104

Of course, the detail must be specified in this file as well.

tck/features/OrderByAcceptance.feature
```diff
@@ -0,0 +1,151 @@
...
+
+Feature: OrderByAcceptance
+
+  Background:
+    Given any graph
+
+  Scenario: ORDER BY should return results in ascending order
+    And having executed: CREATE (n1 {prop: 1}), (n2 {prop: 3}), (n3 {prop: -5})
+    When executing query: MATCH (n) RETURN n.prop AS prop ORDER BY n.prop
+    Then the result should be:
```
Mats-SX: If the query contains `ORDER BY`, the results expectations must match this, by saying `Then the results should be, in order:`, just like here: https://github.com/opencypher/openCypher/blob/master/tck/features/ReturnAcceptance.feature#L33

I see that this is not reflected at all in the readme currently. Please add something about it there as well. There is a section mentioning `ORDER BY`, but it doesn't state this requirement.

tck/features/OrderByAcceptance.feature
```diff
@@ -0,0 +1,151 @@
...
+      | n   |
+      | -5  |
+      | 1   |
+      | 3   |
+
+  Scenario: handle projections with ORDER BY (GH#4937)
+    And having executed: CREATE (c1:Crew {name: 'Neo', rank: 1}), (c2:Crew {name: 'Neo', rank: 2}), (c3:Crew {name: 'Neo', rank: 3}), (c4:Crew {name: 'Neo', rank: 4}), (c5:Crew {name: 'Neo', rank: 5})
+    When executing query: MATCH (c:Crew {name: 'Neo'}) WITH c, 0 AS relevance RETURN c.rank AS rank ORDER BY relevance, c.rank
+    Then the result should be:
+      | rank  |
```
Mats-SX: Many of the results tables here have one space too much in the inner right padding.

petraselmer:
The success of this build depends upon https://github.com/neo4j/neo4j/pull/6891

tck/README.adoc
```diff
@@ -55,7 +55,7 @@ Scenario: Indexing a list with a string
 <2> This step is used to specify an initialization query by scenarios that require certain patterns to exist in the graph.
 <3> If `Q` uses parameters, this step will specify a two-column table detailing the required parameter names and values. The parameter values are in the same format as the expected results.
 <4> The actual Cypher query, `Q`.
-<5> This step specifies the expected results of executing `Q`, represented in a table format if not empty. The first row contains the column names, and subsequent rows contain values. If `Q` contained an `ORDER BY` clause, the result table will be ordered accordingly.
+<5> This step specifies the expected results of executing `Q`, represented in a table format if not empty. The first row contains the column names, and subsequent rows contain values. However, if `Q` contains an `ORDER BY` clause, then this line should instead be `Then the result should be, in order:`, thus ensuring that the result table will be ordered accordingly.
```
Mats-SX: This is not absolutely true. There is an edge case here that makes things a bit more complicated. If the query has `ORDER BY` and at least two columns, then if for some row the values in the ordered column were equal, and the values in at least one other of the columns were not, we can't use `in order` checking of the rows, because any order of those rows is okay.

tck/features/OrderByAcceptance.feature
```diff
@@ -0,0 +1,171 @@
...
+      | limit | -1 |
+    When executing query: MATCH (p:Person) RETURN p.name AS name ORDER BY p.name LIMIT {limit}
+    Then the result should be, in order:
+      | name |
+    And no side effects
+
+  Scenario: ORDER BY with a negative LIMIT should fail with a syntax exception
+    And having executed: CREATE (s:Person {name: 'Steven'}), (c:Person {name: 'Craig'})
+    When executing query: MATCH (p:Person) RETURN p.name AS name ORDER BY p.name LIMIT -1
+    Then a SyntaxError should be raised at compile time: LiteralMustBePositiveInteger
```
Mats-SX: I'm not convinced on the word 'Literal' in the error detail. Do you envision that naming it like this will increase our ability to use this detail in other instances? If not, I'd prefer something like `ArgumentIntegerNegative` or `NegativeIntegerArgument`.

Mats-SX:
The reason the build is failing is due to the error detail not being present in this set: https://github.com/opencypher/openCypher/blob/master/tools/tck/src/main/scala/org/opencypher/tools/tck/constants/TCKErrorTypes.scala#L51

petraselmer:
Merge https://github.com/neo4j/neo4j/pull/6957 after this has been merged

petraselmer:
This PR requires https://github.com/neo4j/neo4j/pull/6958 to have been merged

petraselmer:
Thanks @Mats-SX - I have addressed (hopefully) all of the points you raised.

Mats-SX:
@petraselmer Looks good! Awaiting the implementation PR before merging.

PR #37, opened 2016-04-08
Neatened up the scenario names in the TCK

petraselmer:


Mats-SX:
@petraselmer Nice improvements!

PR #57, opened 2016-05-13
Test grammar using known Cypher queries

Mats-SX:
Also fixes a few whitespace bugs in the grammar.


tools/grammar/src/test/resources/cypher.txt
```diff
@@ -0,0 +1,43 @@
+CREATE ()§
+CREATE( )§
+CYPHER 3.0 MATCH (n) RETURN n§
```
boggle: What are these weird END-OF-LINE symbols? Haven't seen that before

Mats-SX: I needed some non-Cypher character to use as a delimiter between the queries, and used this one.

PR #63, opened 2016-05-18
Port TernaryLogicAcceptance to the TCK

boggle:


tck/features/TernaryLogicAcceptance.feature
```diff
@@ -0,0 +1,160 @@
...
+      RETURN {elt} IN {coll} AS result
+      """
+    Then the result should be:
+      | result   |
+      | <result> |
+    And no side effects
+
+    Examples:
+      | elt  | coll            | result |
+      | null | null            | false  |
```
boggle: This needs a discussion but is true as of now.

tck/features/TernaryLogicAcceptance.feature
```diff
@@ -0,0 +1,160 @@
...
+      | <result> |
+    And no side effects
+
+    Examples:
+      | lhs   | rhs   | result |
+      | null  | null  | null   |
+      | null  | true  | null   |
+      | true  | null  | null   |
+      | null  | false | false  |
+      | false | null  | false  |
```
Mats-SX: Nice! You used the scenario outline feature of Cucumber. I've been thinking about how best to use it; this is an excellent example.

tck/features/TernaryLogicAcceptance.feature
```diff
@@ -0,0 +1,160 @@
...
+    Then the result should be:
+      | value |
+      | null  |
+    And no side effects
+
+  Scenario Outline: Using null in AND
+    And parameters are:
+      | lhs    | <lhs>    |
+      | rhs    | <rhs>    |
+      | result | <result> |
```
Mats-SX: `result` should probably not be a parameter.

tck/features/TernaryLogicAcceptance.feature
```diff
@@ -0,0 +1,160 @@
...
+    And no side effects
+
+    Examples:
+      | elt  | coll            | result |
+      | null | null            | false  |
+      | null | [1, 2, 3]       | null   |
+      | null | [1, 2, 3, null] | null   |
+      | null | []              | false  |
+      | 1    | [1, 2, 3, null] | true   |
+      | 5    | [1, 2, 3, null] | null   |
```
Mats-SX: Perhaps another good example would be `1 IN [null, 1]`, which is `true`, but it makes it clear that even though `null` is encountered first, all elements needs to be processed.

tck/features/TernaryLogicAcceptance.feature
Mats-SX: OK, so given the discussion we had, we should change this to be `null`.

boggle: Let's do that once we have that annotation and keep it as is for now - I've cardified accordingly

Mats-SX: Now we have it: https://github.com/neo4j/neo4j/blob/3.1/community/cypher/compatibility-suite/src/test/scala/cypher/feature/steps/CypherTCKSteps.scala#L43

PR #80, opened 2016-06-02
Port StartPointFindingAcceptance to TCK

Mats-SX:


petraselmer:
I think the name of the file is fairly awkward. I suggest something such as `StartingPointAcceptance` or similar..

Mats-SX:
@petraselmer Renamed the file as per your suggestion.

PR #85, opened 2016-06-03
Style guide improvements

Mats-SX:
- Add new section `Recommendations`
- Add rule for `SKIP` and `LIMIT`
- Add table of contents
- Add recommended file extension `.cypher`

### DIFF
--- a/docs/style-guide.adoc
+++ b/docs/style-guide.adoc
@@ -76,7 +76,7 @@ WHERE n.name CONTAINS 's'
 RETURN n.name
 ----
 
-.. Indent subclauses `ON CREATE` and `ON MATCH` with two spaces.
+.. Indent subclauses with two spaces. Subclauses include `ON CREATE`, `ON MATCH`, `SKIP`, and `LIMIT`.
 +
 .Bad
 [source, cypher]
@@ -85,6 +85,8 @@ MERGE (n) ON CREATE SET n.prop = 0
 MERGE (a:A)-[:T]-(b:B)
 ON CREATE SET a.name = 'me'
 ON MATCH SET b.name = 'you'
+RETURN a.prop SKIP 10
+LIMIT 10
 ----
 +
 .Good
@@ -95,6 +97,9 @@ MERGE (n)
 MERGE (a:A)-[:T]-(b:B)
   ON CREATE SET a.name = 'me'
   ON MATCH SET b.name = 'you'
+RETURN a.prop
+  SKIP 10
+  LIMIT 10
 ----
 
 .. Put `ON CREATE` before `ON MATCH` if both are present.
--- a/docs/style-guide.adoc
+++ b/docs/style-guide.adoc
@@ -6,7 +6,9 @@
 
 toc::[]
 
-Style guide for writing Cypher code.
+This is the style guide for the Cypher language, in the context of its standardization through the openCypher project.
+In this document you will find rules for how to use the language, intended to help construct a consistent and portable usage of the language across many use cases and implementations.
+There are also more peripheral recommendations, which are to be taken more lightly, and reflects attempts at helping users of the language to treat language-related concerns in a consistent manner.
 
 NOTE: Cypher is a language.
 Languages have their own rules, and are not eager to adapt to circumstance.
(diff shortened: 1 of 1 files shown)


docs/style-guide.adoc
```diff
@@ -1,6 +1,14 @@
...
+:toc:
+:toc-placement: macro
+
 = The Cypher Style Guide
 
-Style guide for writing Cypher code.
+toc::[]
+
+This is the style guide for the Cypher language, in the context of its standardization through the openCypher project.
+In this document you will find rules for how to use the language, intended to help construct a consistent and portable usage of the language across many use cases and implementations.
```
petraselmer: "In this document you will find rules for how to use the language" -> "This document contains rules for how to use the language"

docs/style-guide.adoc
```diff
@@ -35,7 +43,7 @@ ORDER BY score DESC
 
 == Rules
 
-Whenever two rules are in conflict, use the order to decide which one trumps.
+In case two rules are in conflict, and there is no explicit mention which rule trumps, the last mentioned rule applies.
```
petraselmer: "In case two rules are in conflict, and there is no explicit mention which rule trumps, the last mentioned rule applies" -> "In case two rules are in conflict, and there is no explicit mention of which rule trumps, the rule mentioned last applies"

docs/style-guide.adoc
```diff
@@ -1,6 +1,14 @@
...
+:toc-placement: macro
+
 = The Cypher Style Guide
 
-Style guide for writing Cypher code.
+toc::[]
+
+This is the style guide for the Cypher language, in the context of its standardization through the openCypher project.
+In this document you will find rules for how to use the language, intended to help construct a consistent and portable usage of the language across many use cases and implementations.
+There are also more peripheral recommendations, which are to be taken more lightly, and reflects attempts at helping users of the language to treat language-related concerns in a consistent manner.
```
petraselmer: I would also change the heading "== Why Is Style Important?" -> "== Why is style important?"

docs/style-guide.adoc
```diff
@@ -1,6 +1,14 @@
+:numbered:
```
petraselmer: "Spacing" and "patterns" (paragraphs numbered 9 and 13) are both in bold, in contrast to all of the other headings. Either all should be bold (my preference) or none should be bold

docs/style-guide.adoc
Mats-SX: But `more` is used as strengthening `peripheral` here; `some` wouldn't do that. I don't know how to rephrase to keep the intent.

docs/style-guide.adoc
Mats-SX: It's bold if it is a category of rules, rather than a specific rule.

petraselmer: It does look a bit awkward, though...and they're both very long sections as well. Maybe they warrant being "pulled out", just like "Graph modelling" is for the Recommendations section?

docs/style-guide.adoc
```diff
@@ -526,3 +523,28 @@ RETURN 1;
 ----
 RETURN 1
 ----
+
+== Recommendations
+
+* When using Cypher language constructs in prose, use a monospaced font and follow the styling rules.
+** When referring to labels and relationship types, include the colon like so: `:Label`, `:REL_TYPE`.
```
petraselmer: "When referring to labels and relationship types, include the colon like so:" -> "When referring to node labels and relationship types, the colon should be included as follows:"

docs/style-guide.adoc
```diff
@@ -526,3 +523,28 @@ RETURN 1;
...
+** When referring to functions, use lower camel case and parentheses like so: `shortestPath()`. Arguments should normally not be included.
```
petraselmer: "...use lower camel case and parentheses like so:" -> "...lower camel case and parentheses should be used as follows:"

petraselmer:
@Mats-SX I have added a few more comments

docs/style-guide.adoc
Mats-SX: That's a good idea, I'll try that.

Mats-SX:
@petraselmer Made changes according to comments. WDYT about the new sections?

petraselmer:
@Mats-SX the new sections look really good, as do the other changes - much more readable! Thanks - and merging...

PR #88, opened 2016-06-08
Ensure elements in list are ordered so that test is stable

petraselmer:
Merging as @systay and I paired on this


Mats-SX:
This could've been a great use case for the unordered matching of lists :)

Mats-SX:
@petraselmer When using `ORDER BY`, we should use the step `Then the results should be, in order`.

PR #89, opened 2016-06-09
Fixes for named graphs

Mats-SX:


tck/graphs/named-graphs.adoc
```diff
@@ -67,3 +67,8 @@ For label/sublabel combinations, the supported advice is:
 
 * `implies`
 ** This indicates that the existence of the label on a node implies the existence of the sublabel. In other words, if a node has the label, it will always have the sublabel.
+
+=== Scripts
+
+The metadata file will specify one or more script files, which are files that contain Cypher statements which are to be used to create the named graph.
```
petraselmer: "The metadata file will specify one or more script files, which are files that contain Cypher statements which are to be used to create the named graph." 
->
"The metadata file will specify one or more script files that contain Cypher statements which are used to create the named graph."

Also (from higher up in the file):
"For some statistical combinations, the metadata file an optional piece of" -> "For some statistical combinations, the metadata file contains an optional piece of" 

"This is called 'advice'" - is there a better term we can use? This sits very oddly here and later on in the file. Would one of "direction, directive, guide, suggestion or hint" be amenable?

Mats-SX: I'll fix the two grammar things. On advice, I'm inclined to perhaps leave it out altogether, if not now perhaps later, so I don't want to spend too much thinking on it.

PR #97, opened 2016-06-20
Redefine comparability and equality, as well as orderability and equivalence

boggle:


cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+:toc:
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Author:* Mats Rydberg <mats@neotechnology.com>, Stefan Plantikow <stefan.plantikow@neotechnology.com>
+
+[abstract]
+.Abstract
+--
+This CIP intends to introduce and formalise two new concepts, *comparability* and *orderability*, and defines their relationship to the already existing concepts of *equality* and *equivalence*.
```
petraselmer: "already existing" -> "pre-existing"
I see all 4 concepts are covered in this CIP... so was wondering how `equality` and `equivalence` are pre-existing? Are they in another CIP or defined somewhere else? If yes, this other resource ought to be linked here...

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+Comparability revolves around the inequality operators (`>`, `<`, `>=`, `<=`), and defines the semantics they need to properly determine how two values compare to each other.
```
petraselmer: This - and operability - are good definitions. I think, though, the abstract may not be best-placed for these definitions. I would just mention here that all 4 concepts are harmonized and defined more precisely (as well as your "in summary" sentence), and then have a `Definitions` section as the first sub-section under `Proposal`.
This is to aid the 'flow' of the document

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+Futhermore, the definitions of *equality* and *equivalence* are harmonized and spelled out.
+In summary, we are proposing some changes to how Cypher defines these four concepts in order to get a consistent set of rules and a simpler conceptual model.
+--
+
+toc::[]
+
+
+== Motivation
+
+There are currently a number of limitations and inconsistencies that this CIP aims to fix.
```
petraselmer: Please can it be spelled out precisely where these limitations are (I am guessing it is in Cypher itself)? Otherwise, this may be a tad unclear to readers :-)

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+(1) Cypher already has good semantics for equality within the primitive types (booleans, strings, integers, and floats) and maps.
+Furthermore, Cypher has good semantics for comparability and orderability for integers, floats, and strings, within the types.
```
petraselmer: "within the types" -> "within each of the types"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+However working with values of different types can be difficult:
+
+  * Comparability between values of different types is often undefined, which is something we regard as a problem for certain combinations of values and types.
+  * `ORDER BY` will often fail with an error if the values passed to it have different types.
+
+(2) The underlying conceptual model is complex and sometimes inconsistent which leads to an unclear relationship between comparison operators, equality, grouping, and `ORDER BY`:
```
petraselmer: "The underlying conceptual model is complex and sometimes inconsistent which leads to an unclear relationship between comparison operators, equality, grouping, and `ORDER BY`:" 
--->
"The underlying conceptual model is complex and sometimes inconsistent.  This leads to an unclear relationship between comparison operators, equality, grouping, and `ORDER BY`:"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
petraselmer: I would also add that, in addition to the limitations etc, another motivation for this CIP is to unambiguously and precisely define these 4 concepts.

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+  * Comparability and orderability are not aligned with each other consistently (Some types may be ordered but not compared).
```
petraselmer: Is the content in () the only inconsistency? If not, I would suggest "Comparability and orderability are not aligned with each other consistently, as some types may be ordered but not compared."

If it is not, I would suggest "Comparability and orderability are not aligned with each other consistently, such as the fact that some types may be ordered but not compared."

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+  * There are various inconsistencies around equality (and equivalence) semantics as exposed by `IN`, `=`, `DISTINCT`, and grouping.
+  The difference between equality and equivalence in Cypher today is small and subtle, and limited to testing two instances of the value `null` to each other.
+
+    ** In equality, `null = null` is `null`.
+    ** In equivalence, used by grouping and `DISTINCT`, instances of `null` are always treated as being the same value.
```
petraselmer: ", used by grouping and `DISTINCT`, " --> " - used by `DISTINCT` and when grouping values - "

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+    ** However, equality treats `null` values differently if they are an element of a list or a map value.
+    ** Similar rules apply for `NaN`.
+
+== Proposal
+
```
petraselmer: I would suggest adding your defs from the abstract to a new 'Definitions' section here, and also add defs for equality and equivalence, along similar lines to what you did for comparability and orderability

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+=== Comparability
+
+We propose that comparability be defined between any pair of values, as specified below (Please note that `null` is interpreted as meaning "unknown" or "could be any value" in the context of comparability):
```
petraselmer: "Please" --> "please"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+* Integers are compared numerically.
+* Floats (excluding `NaN` and the Infinities) are compared numerically.
+* Arbitrary numbers (excluding `NaN` and the Infinities) are compared to each other numerically, i.e. the comparison behaves as if both numbers would have been coerced to arbitrary precision big decimals (currently outside the Cypher type system) before comparing them with each other numerically.
```
petraselmer: Given that integers and floats have been mentioned already, what precisely is an "arbitrary" number?

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* Positive infinity is of type `FLOAT`, equal to itself and larger than any other number (excluding `NaN`).
+* Negative infinity is of type `FLOAT`, equal to itself and smaller than any other number (excluding `NaN`).
+* Comparing a number to a `NaN` always yields `null`.
+* Booleans are compared such that `false < true`.
+* Strings are compared in dictionary order, i.e. characters are compared pairwise in order and characters missing in a shorter string are considered to be smaller than any other character. For example, `'a' < 'aa'`.
+* Lists are compared in dictionary order, i.e. list elements are compared pairwise in order and elements missing in a shorter list are considered to be smaller than any other value. For example, `[1] < [1,0]`.
+* Lists containing `null` values are incomparable, i.e. comparing them with any other value yields `null`.
```
petraselmer: To make it even clearer, can an example be provided (just as you have done for the points above?)

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* Maps are compared as if they where lists of lists, where the inner lists are the key-value pairs of the maps, ordered in ascending sort order for the keys. Comparing a map to a value that is not a map always yields `null`.
```
petraselmer: "if they where" --> "if they were". 

An example here would be beneficial...

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+* Integers are compared numerically.
+* Floats (excluding `NaN` and the Infinities) are compared numerically.
+* Arbitrary numbers (excluding `NaN` and the Infinities) are compared to each other numerically, i.e. the comparison behaves as if both numbers would have been coerced to arbitrary precision big decimals (currently outside the Cypher type system) before comparing them with each other numerically.
+* Positive infinity is of type `FLOAT`, equal to itself and larger than any other number (excluding `NaN`).
+* Negative infinity is of type `FLOAT`, equal to itself and smaller than any other number (excluding `NaN`).
+* Comparing a number to a `NaN` always yields `null`.
+* Booleans are compared such that `false < true`.
+* Strings are compared in dictionary order, i.e. characters are compared pairwise in order and characters missing in a shorter string are considered to be smaller than any other character. For example, `'a' < 'aa'`.
+* Lists are compared in dictionary order, i.e. list elements are compared pairwise in order and elements missing in a shorter list are considered to be smaller than any other value. For example, `[1] < [1,0]`.
```
petraselmer: Stylistically, could "List" not have sub bullets, so as to draw all the points closer together (than having the info spread out across 2 bullets, as it is at present)?

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* Lists containing `null` values are incomparable, i.e. comparing them with any other value yields `null`.
+* Maps are compared as if they where lists of lists, where the inner lists are the key-value pairs of the maps, ordered in ascending sort order for the keys. Comparing a map to a value that is not a map always yields `null`.
+* Maps containing explicit `null` value entries are incomparable, i.e. comparing them with any other value yields `null`.
```
petraselmer: Just as for "Lists", could "Map" not have sub bullets, so as to draw all the points closer together (than having the info spread out across 2 bullets, as it is at present)?

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* The comparison order for nodes (and relationships resp.) is undefined, and an implementation may choose an order for these freely, e.g. by using some kind of id. This order must not change during the execution of the query (i.e. it must be stable in the presence of updates). Comparing a node to a relationship always yields `null`.
```
petraselmer: "some kind of id" -> this is rather hand-wavey and fuzzy. Can this be made more precise? At the very least, please replace "kind" with "sort"...

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* Paths are compared as if they where a list of alternating nodes and relationships of the path from the start node to the end node. Comparing a path to a value that is not a path always yields `null`.
```
petraselmer: "they where a list" --> "they were a list"

petraselmer: The above description does not say, however, precisely how the paths are compared. Is it numeric or dictionary order of the composite ids? What about length? Examples would be very beneficial too.

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+* `NODE`
+* `RELATIONSHIP`
+* `PATH`
+* `MAP`
+* `LIST OF ANY?`
+* `STRING`
+* `BOOLEAN`
+* `NUMBER` (`NaN` is treated as the largest number in orderability only, i.e. it is put after positive infinity)
+* `VOID` (i.e. the type of `null`)
```
petraselmer: The type order is a very important concept, albeit not the simplest (it is very subtle!), and in light of this, I think that an example here would be a very good thing. Something along the lines of if two values, v1 and v2 were ordered, and v1 was a NODE whereas v2 was a MAP, v1 would be ordered before v2? Otherwise, if a misunderstanding occurs here, the reader is lost...

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+Within the types, orderability defers to comparability with two exceptions:
```
petraselmer: Neither of the asterisks (*) appear to be formatted: these just run and wrap on the same line. Can these be pulled out into sub bullets?

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* Two `null` values are treated as the same value in accordance with the global sort order
+* Two `NaN` values are treated as the same value in accordance with the global sort order
+
+The accompanying descending global sort oder is the same order in reverse (i.e. it runs from `VOID` to `NODE`).
+
+Any additionally introduced non-canonical types introduced by an implementation should be inserted between `PATH` and `MAP` in the global sort order. As an exception, additionally introduced non-canonical number types may be integrated under the `NUMBER` type in the global sort order.
```
petraselmer: "Any additionally introduced non-canonical types introduced by" --> 
"Any additional non-canonical types introduced by"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* `STRING`
+* `BOOLEAN`
+* `NUMBER` (`NaN` is treated as the largest number in orderability only, i.e. it is put after positive infinity)
+* `VOID` (i.e. the type of `null`)
+
+Within the types, orderability defers to comparability with two exceptions:
+* Two `null` values are treated as the same value in accordance with the global sort order
+* Two `NaN` values are treated as the same value in accordance with the global sort order
+
+The accompanying descending global sort oder is the same order in reverse (i.e. it runs from `VOID` to `NODE`).
```
petraselmer: "global sort oder" -> "global sort order"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+Any additionally introduced non-canonical types introduced by an implementation should be inserted between `PATH` and `MAP` in the global sort order. As an exception, additionally introduced non-canonical number types may be integrated under the `NUMBER` type in the global sort order.
+
+=== Equality ===
+
+We propose that equality and comparability must be aligned, i.e. `l = r` if and only if `l <= r && l >= r`.
```
petraselmer: "that equality and comparability must be aligned" 
--->
"that equality and comparability be aligned"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
petraselmer: "As an exception, additionally introduced non-canonical number types" 
--->
"As an exception, any additional non-canonical number types"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
petraselmer: A reminder that '<=' renders oddly

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+To achieve this, we unify the difference between equality and equivalence, by making lists containing `null` values follow the same difference as the `null` values themselves.
```
petraselmer: "To achieve this, we unify the difference between equality and equivalence, by making" 
--->
"We thus unify the difference between equality and equivalence by making"

petraselmer: "values follow the same difference as the `null` values themselves" - this snippet does not make sense. What does 'follow the same difference as the `null` values' mean?

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+Concretely, we propose to redefine how equality works for lists in Cypher today.
```
petraselmer: "Concretely" --> "Specifically"
I would also drop "today"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+To determine if two lists `l1` and `l2` are equal, we propose two simple tests, like so
```
petraselmer: "we propose two simple tests, like so"
-->
"we propose two simple tests, as exemplified by the following: "

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+* `l1` and `l2` must have the same size, i.e. inversely `size(l1) <> size(l2>) => l1 <> l2`
+* the pairwise elements of both `l1` and `l2` must be equal, i.e.
+----
+[a1, a2, ...] = [b1, b2, ...]
+<=>
+a1 = b1 && a2 = b2 && ...
+----
```
petraselmer: I like the above example very much 👍

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+For clarity, we also repeat the current equality semantics of maps here (at the time of writing this CIP). Two maps `m1` and `m2` are considered equal if
```
petraselmer: I would drop "(at the time of writing this CIP)"

petraselmer: "considered equal if" --> "considered equal if:"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+* `m1` and `m2` must have the same keys.
```
petraselmer: "`m1` and `m2` must have the same keys." --> 
"`m1` and `m2` have the same keys, and"

(this rewording follows on from the condition set prior to the bullets)

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* For each key `k`,
+** either `m1.k = m2.k` is `true`,
+** or both `m1.k IS NULL` and `m2.k IS NULL`
+
+This is at odds with the common interpretation of `null` as standing for any possible value.
```
petraselmer: "standing for" --> "representing"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+It is aligned though with the most common use case for maps with `null` entries which is to update multiple properties at once, e.g. `SET n += { n: 12, remove_this_key: null }`. In this case, there is no need to differentiate between different `null` values as `null` merely serves as a marker for keys to be removed. Current equality semantics makes it easy to check if two maps correspond to the same property update in this scenario.
```
petraselmer: "It is aligned though with the most common use case"
-->
"However, our interpretation of `null` is aligned with the most common use case"

petraselmer: "to update multiple properties at once" 
-->
"to update multiple properties through the use of a single clause" (or similar)

petraselmer: Is `SET n += { n: 12...` correct? Surely `n` is the node var, and thus not able to be used inside the property map... so something like `SET n += { age: 12` ?

petraselmer: "different `null` values as `null` merely"
--->
"different `null` values, as `null` merely"

petraselmer: "Current equality semantics makes it" 
-->
"Current equality semantics make it"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+However this type of update map comparison is rare and could be emulated using a more complex predicate. The current rules do however break symmetry with how equality handles `null` in all other cases. This becomes more apparent by considering these two examples
```
petraselmer: "However this type" 
--->
"We note that this type" (avoids overuse of 'however')

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+* `expr1 = expr2` => `null` if `expr1 IS null && expr2 IS NULL`
+* `{a: expr1} = {a: expr2}` => `true` (!!) if `expr1 IS null && expr2 IS NULL`
+
+To fix this, we propose that two maps `m1` and `m2` instead should be equal if
```
petraselmer: "we propose that two maps `m1` and `m2` instead"
--->
"we propose instead that two maps `m1` and `m2`"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
petraselmer: I would emphasize 'current' here - this is easily lost at the moment, and weakens the motivation for the rest of the discourse on maps...

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
petraselmer: "be equal if"
--->
"be equal if: "

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+* `m1` and `m2` have the same keys (the order of keys as returned by `keys` does not matter)
```
petraselmer: "does not matter)"
--->
"does not matter), and"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* For each such key `k`, `m1.k = m2.k`
+
```
petraselmer: Please add something indicating that this implies neither `m1.k` nor `m2.k` is `null` (perhaps in brackets). This is to make it crystal clear to the reader, even though this is already implied by the statement

petraselmer: Also, ought "is `true`" not be appended to the sentence?

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+** or both `m1.k IS NULL` and `m2.k IS NULL`
+
+This is at odds with the common interpretation of `null` as standing for any possible value.
+
+It is aligned though with the most common use case for maps with `null` entries which is to update multiple properties at once, e.g. `SET n += { n: 12, remove_this_key: null }`. In this case, there is no need to differentiate between different `null` values as `null` merely serves as a marker for keys to be removed. Current equality semantics makes it easy to check if two maps correspond to the same property update in this scenario.
+
+However this type of update map comparison is rare and could be emulated using a more complex predicate. The current rules do however break symmetry with how equality handles `null` in all other cases. This becomes more apparent by considering these two examples
+
+* `expr1 = expr2` => `null` if `expr1 IS null && expr2 IS NULL`
+* `{a: expr1} = {a: expr2}` => `true` (!!) if `expr1 IS null && expr2 IS NULL`
```
petraselmer: Can something more elegant than "(!!)" be used to point out the contradiction/problem here? Perhaps some text in brackets pointing out the problem?

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+To fix this, we propose that two maps `m1` and `m2` instead should be equal if
+
+* `m1` and `m2` have the same keys (the order of keys as returned by `keys` does not matter)
+* For each such key `k`, `m1.k = m2.k`
+
+As a consequence of these changes, plain equality is not reflexive in a classic sense (Consider: `{a: null} = {a: null}`, `[null] = [null]`). However this was already the case (Consider: `null = null` => `null`).
```
petraselmer: "(Consider" ---> "(consider"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+However, Equality is reflexive for values that do not involve `null`.
```
petraselmer: "Equality" --> "equality"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+=== Equivalence ===
+
+Equivalence remains unchanged but now can be defined succinctly as being identical to equality except that:
```
petraselmer: Once more, I think the next two asterisked points have no formatting.... these ought to be bullet points

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* Any two `null` values are treated as equivalent (both directly or inside nested structures).
+* Any two `NaN` values are treated as equivalent (both directly or inside nested structures).
+
+Note that orderability is aligned with equivalence, i.e. equivalent values have the same position under orderability.
+
+Equivalence is reflexive for all values.
+
+=== Summary of Conceptual Model
```
petraselmer: "Summary of Conceptual Model" --> "Summary of the conceptual model"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+This proposal aims to simplify the conceptual model around equality, comparison, sorting, and grouping:
+
+* Equality: Equality follows natural, literal equality. However, values involving `null` are never equal. Nested structures are first tested for equality by shape (keys, size) and then componentwise. This ensures that equality is compatible with interpreting `null` as "unknown" or "could be any value".
```
petraselmer: "componentwise"? Can this be explained in a clearer way?

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+* Comparability: Comparability is aligned with equality. Two values of the same type in the global sort order are always comparable. Two values of different types are always incomparable and values involving `null` are always incomparable.
```
petraselmer: It would be nice if "global sort order" could link back to that section in the document in which this was defined

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+This ensures that `MATCH (n) WHERE n.prop > 42` will never find nodes where `n.prop` is of type `STRING`.
+* Equivalence: Equivalence is a form of equality that treats `null` (and `NaN`) values as identical. Equivalence is used in grouping and `DISTINCT` where `null` commonly is interpreted as a marker that a value was missing instead of as a wildcard for any possible value.
+* Orderability follows comparability but additionally defines a global sort order between values of different types and is aligned with equivalence regarding the handling of `null` and `NaN` values.
```
petraselmer: As per my comment above regarding "global sort order"

petraselmer: "and is aligned with equivalence regarding" - this is fine, but even so, I would add an "i.e. <re-explain how this works even if repeating from what was said about `null` and `NaN` in the equivalence point>" at the end of the sentence (for readability and understanding)

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+=== Examples
```
petraselmer: This section is great, and I think many readers will use this as a reference point to validate their understanding of the text. In light of that, I strongly suggest adding more examples.

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+=== Interaction with existing features
+
+The concept of orderability is used only by `ORDER BY` in Cypher today.
+The concept of comparability is used by the comparisons operators `<`, `>`, `<=`, >=`.
+
+One major goal of our proposal is for equality semantics to align well with comparability.
+The concept of equality is used by the equality operator `=`, the inequality operator `<>`, value joins, and the `IN` operator.
+The concept of equivalence is used by the `DISTINCT` clause modifier and in grouping.
+
+With the proposals made in this CIP, specifically changing equality for lists, the mentioned functionality is going to treat lists containing `null` as unequal, thus potentially filtering out more rows when used in a predicate.
```
petraselmer: This is a very, very, very long sentence. 

"Owing to this proposal altering equality semantics for lists - in particular, treating lists containing `null` as unequal - potentially more rows may be filtered out when used in a predicate." or something along these lines...

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+=== Alternatives
+
+Columns in SQL are always have a concrete type. This removes the need to define a global sort order between types. Standard SQL has no support for lists, maps, or graph structures and hence does not need to define semantics for them.
```
petraselmer: "Columns in SQL are always have a concrete type" 
-->
"Columns in SQL always have a concrete type"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+SQL also treats comparisons involving `null` as returning `null`.
+
+PostgresSQL treats some numerical operations (like division by zero) that would compute a `NaN` as a numerical error that fails the query. PostgresQL considers `NaN` to be larger than positive infinity, both in comparison and in sort order. This proposal achieves something very similar by evaluating comparisons involving a `NaN` to `null` and by treating both `NaN` and `null` as the largest values in the global sort order.
```
petraselmer: "(like division" --> "(such as division"

petraselmer: "error that fails the query." 
--->
"error that causes the query to fail."

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+This proposal could be extended with an operator for making equivalence accessible beyond use in grouping and `DISTINCT`. This seems desirable due to plain `=` not being reflexive for all values.
+
+Also noteworthy: If nulls would track their source, equality could become reflexive again as it would become possible to know if two `null` values represent the same "unknown" value.
+
+== Benefits to this proposal
+
+A consistent set of rules for equality, equivalence, comparability and orderability.
```
petraselmer: "A consistent set of rules for equality, equivalence, comparability and orderability." 
-->
"A consistent set of rules is defined for equality, equivalence, comparability and orderability."
(Currently a clause rather than a sentence...)

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+== Caveats to this proposal
+
+Adopting this proposal may break some queries; specifically queries that depend on equality semantics of lists containing `null` values.
+It should be noted that we expect that most lists used in queries are constructed using `collect()`, which never outputs `null` values.
+
+== Appendix: Comparability by Type
+
+The following table captures which types may be compared with each other such that the outcome is either `true` or `false`.
+Any other comparison will always yield `null` (except for `NaN`) which is handled as described above.
+
```
petraselmer: 👍  👍

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
Mats-SX: They exist de facto, but I don't know of any specification document. They do need to be defined, I agree.

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
Mats-SX: The asciidoc just needs a newline to make the bullet points work.

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
```diff
@@ -0,0 +1,233 @@
...
+*Author:* Mats Rydberg <mats@neotechnology.com>, Stefan Plantikow <stefan.plantikow@neotechnology.com>
+
+[abstract]
+.Abstract
+--
+This CIP intends to introduce and formalise two new concepts, *comparability* and *orderability*, and defines their relationship to the already existing concepts of *equality* and *equivalence*.
+Comparability revolves around the inequality operators (`>`, `<`, `>=`, `<=`), and defines the semantics they need to properly determine how two values compare to each other.
+Orderability revolves around the `ORDER BY` clause, and defines the semantics it needs to be able to properly sort values passed to it.
+Futhermore, the definitions of *equality* and *equivalence* are harmonized and spelled out.
```
petraselmer: "Futhermore" -> "Furthermore"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
boggle: "including keys mapping to `null`"

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
boggle: -- INCONSISTENT!

cip/CIP2016-06-14-Harmonize-comparability-orderability-equality-and-equivalence.adoc
boggle: This was a great find! It lead to the discovery of a forgotten corner case.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+:toc:
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Author:* Mats Rydberg <mats@neotechnology.com>, Stefan Plantikow <stefan.plantikow@neotechnology.com>
+
+[abstract]
+.Abstract
+--
+This CIP redefines and formalizes 4 key lanugage concepts, *comparability* and *equality*, as well as *orderability* and *equivalence*. While these notions already exist in the language today, they have never been defined explicitly.
```
Mats-SX: typo: `lanugage`

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+[[comparability-con,comparability (concept)]]*Comparability*:: <<comparability-def,Comparability>> revolves around the inequality operators (`>`, `<`, `>=`, `\<=`), and defines the semantics they need to properly determine how two values compare to each other.
+[[equality-con,equality (concept)]]*Equality*:: <<equality-def,Equality>> revolves around the equality operators ('=', '<>') as well as the list membership operator (`IN`) and value joins, and defines the semantics the needed to properly determine if two values are the same.
+Equality is also used implicitly by literal maps in node and relationship patterns, since such literal maps are merely a shorthand notation for equality predicates.
+[[orderability-con,orderability (concept)]]*Orderability*:: <<orderability-def,Orderability>> revolves around the `ORDER BY` clause, and defines the semantics it needs to be able to properly sort values passed to it.
+[[equivalence-con,equivalence (concept)]]*Equivalence*:: <<equivalence-def,Equivalence>> revolves around the `DISTINCT` modifier and grouping in projection clauses (`WITH`, `RETURN`), and defines the semantics they need to properly determine if two values are the same.
```
Mats-SX: I propose `includes` instead of `revolves around` for the four concepts.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+We propose to redefine <<comparability-def>> and <<equality-def>>, as well as <<orderability-def>> and <<equivalence-def>> as detailed in this section.
+
+[[concepts]]
+=== Concepts
+
+Cypher today features four distinct concepts related to equality and ordering:
+
+[horizontal]
+[[comparability-con,comparability (concept)]]*Comparability*:: <<comparability-def,Comparability>> revolves around the inequality operators (`>`, `<`, `>=`, `\<=`), and defines the semantics they need to properly determine how two values compare to each other.
+[[equality-con,equality (concept)]]*Equality*:: <<equality-def,Equality>> revolves around the equality operators ('=', '<>') as well as the list membership operator (`IN`) and value joins, and defines the semantics the needed to properly determine if two values are the same.
```
Mats-SX: Use backticks instead of apostrophes for operator symbols (=, <>).

Mats-SX: I propose rephrase `Equality ... the equality operators, the list membership operator, and value joins`; replace `as well as` with a comma.

Mats-SX: `the needed` -> `they need`

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+[[equivalence-con,equivalence (concept)]]*Equivalence*:: <<equivalence-def,Equivalence>> revolves around the `DISTINCT` modifier and grouping in projection clauses (`WITH`, `RETURN`), and defines the semantics they need to properly determine if two values are the same.
+
+==== The meaning of `null`
+
+For the following discussion, it is helpful to clarify the meaning of `null`. In Cypher, a `null` value has one of two meanings, depending on the context in which it occurs:
+
+[horizontal]
+[[unknown-null,unknown `null`]]*Unknown*:: An "unknown" `null` is taken to be a placefolder for an arbitrary but unknown value. When evaluating predicates, an "unknown" `null` is the `maybe` truth value of ternary logic. For node and relationship properties, an "unknown" `null` is a value that is definite in the real world but has not been stored in the graph. Since in these cases, two "unknown" `null` values stand for arbitrary but definite values in the real world, two "unknown" `null` values should never be treated as certainly being the same value.
```
Mats-SX: typo: `placefolder`

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+[[equality-con,equality (concept)]]*Equality*:: <<equality-def,Equality>> revolves around the equality operators ('=', '<>') as well as the list membership operator (`IN`) and value joins, and defines the semantics the needed to properly determine if two values are the same.
+Equality is also used implicitly by literal maps in node and relationship patterns, since such literal maps are merely a shorthand notation for equality predicates.
+[[orderability-con,orderability (concept)]]*Orderability*:: <<orderability-def,Orderability>> revolves around the `ORDER BY` clause, and defines the semantics it needs to be able to properly sort values passed to it.
+[[equivalence-con,equivalence (concept)]]*Equivalence*:: <<equivalence-def,Equivalence>> revolves around the `DISTINCT` modifier and grouping in projection clauses (`WITH`, `RETURN`), and defines the semantics they need to properly determine if two values are the same.
+
+==== The meaning of `null`
```
Mats-SX: Love this section!

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+[[missing-null,missing `null`]]*Missing*:: A "missing" `null` is taken to be a marker for the absence of a value. In the context of updating node properties from a map, a "missing" `null` is used to mark properties that are to be removed. In the context of `DISTINCT` and grouping, a "missing" `null` value is used as grouping key for all records that miss a more specific value. Since in these cases, two "missing" `null` values represent the same concept, they should always be treated as the same value.
+
+[[regular-maps]]
+==== Regular maps
+
+Cypher today has one supertype `MAP` for all map values. This includes nodes (of subtype `NODE`), relationships (of subtype `RELATIONSHIP`), and any other map (not captured by a subtype of `MAP`). For the purpose of this document, we define a regular map to be any value of type `MAP` that is neither a `NODE` nor a `RELATIONSHIP`.
+
+<<comparability-equality,comparability and equality>>
```
Mats-SX: This asciidoc doesn't render correctly. I think using just `<<comparability-equality>>` should be fine.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+=== Comparability and equality
+
+We propose that comparability and equality should be aligned with each other, i.e.
+
+`expr1 = expr2` if and only if `expr1 >= expr2 && expr1 \<= expr2`.
+
+Comparability and equality use <<unknown-null,"unknown" `null` values>>.
```
Mats-SX: These links don't seem to work. I'm not sure asciidoc links should have two arguments, but only one (the key, which should try to avoid spaces).

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+
+[[incomparable,incomparable]]
+==== Incomparability
+
+We define that a value is incomparable if and only if any comparison or equality test involving this value evaluates to `null`.
```
Mats-SX: What is the purpose of this definition? It seems to define every single value as incomparable, seeing as comparing any value to `null` will evaluate to `null`. If it is motivated, please provide a motivation.

Mats-SX: I think that perhaps the definition should be `if and only if _all_ comparisons and equality tests ... evaluates to null`, in which case I retract my previous comment.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+
+We propose that <<comparability-con,comparability>> should be defined between any pair of values, as specified below.
+
+- Numbers
+  * Integers are compared numerically in ascending order.
+  * Floats (excluding `NaN` and the Infinities) are compared numerically in ascending order.
+  * Numbers of different types (excluding `NaN` and the Infinities) are compared to each other as if both numbers would have been coerced to arbitrary precision big decimals (currently outside the Cypher type system) before comparing them with each other numerically in ascending order.
+  * Positive infinity is of type `FLOAT`, equal to itself and larger than any other number (excluding `NaN`).
+  * Negative infinity is of type `FLOAT`, equal to itself and smaller than any other number (excluding `NaN`).
+  * `NaN` is <<incomparable>>.
```
Mats-SX: Hah! This one makes me feel good.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+  * Numbers are <<incomparable>> to any value that is not also a number.
+- Booleans
+  * Booleans are compared such that `false < true`.
```
Mats-SX: This is either a recursive definition with no base case, or relying on an implicit overload of the `<` operator in the reader's mind; I propose we spell it out in English: `... such that false is less than true`.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+  * Booleans are <<incomparable>> to any value that is not also a boolean.
+- Strings
+  * Strings are compared in dictionary order, i.e. characters are compared pairwise in ascending order from the start of the string to the end. Characters missing in a shorter string are considered to be smaller than any other character. For example, `'a' < 'aa'`.
+  * Strings are <<incomparable>> to any value that is not also a string.
+- Lists
+  * Lists are compared in dictionary order, i.e. list elements are compared pairwise in ascending order from the start of the list to the end. Elements missing in a shorter list are considered to be smaller than any other value. For example, `[1] < [1,0]`.
```
Mats-SX: Missing space after comma in the list.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+  * Lists containing `null` values are <<incomparable>>. For example, `[1] >= [1, null]` evaluates to `null`.
+  * Lists are <<incomparable>> to any value that is not also a list.
+- Maps
+  * [[regular-maps,regular maps]]Regular maps
+  ** The comparison order for maps is unspecified and left to implementations.
+  ** The comparison order for maps must align with the <<equality-def,equality semantics>> outlined below. In consequence, any map that contains an entry that maps it's key to a `null` value is <<incomparable>>. For exampe, `{a: 1} \<= {a: 1, b: null}` evaluates to `null`.
```
Mats-SX: `it's` -> `its`

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+  ** Regular maps are <<incomparable>> to any value that is not also a regular map.
```
Mats-SX: Is it [this](https://en.wikipedia.org/wiki/Regular_map_(graph_theory)) kind of regular map that is referred to? Is that a well-known concept?

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+- Strings
+  * Strings are compared in dictionary order, i.e. characters are compared pairwise in ascending order from the start of the string to the end. Characters missing in a shorter string are considered to be smaller than any other character. For example, `'a' < 'aa'`.
+  * Strings are <<incomparable>> to any value that is not also a string.
+- Lists
+  * Lists are compared in dictionary order, i.e. list elements are compared pairwise in ascending order from the start of the list to the end. Elements missing in a shorter list are considered to be smaller than any other value. For example, `[1] < [1,0]`.
+  * Lists containing `null` values are <<incomparable>>. For example, `[1] >= [1, null]` evaluates to `null`.
+  * Lists are <<incomparable>> to any value that is not also a list.
+- Maps
+  * [[regular-maps,regular maps]]Regular maps
+  ** The comparison order for maps is unspecified and left to implementations.
```
Mats-SX: I'd like to have Tobias' definition spelled out here as well, namely that _equal_ values are grouped together.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+  ** The comparison order for maps must align with the <<equality-def,equality semantics>> outlined below. In consequence, any map that contains an entry that maps it's key to a `null` value is <<incomparable>>. For exampe, `{a: 1} \<= {a: 1, b: null}` evaluates to `null`.
+  ** Regular maps are <<incomparable>> to any value that is not also a regular map.
+  * Nodes
+  ** The comparison order for nodes is unspecified and left to implementations.
```
Mats-SX: I'd like to have Tobias' definition spelled out here as well, namely that _equal_ values are grouped together.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+  ** Nodes are <<incomparable>> to any value that is not also a node.
+  * Relationships
+  ** The comparison order for relationships is unspecified and left to implementations.
```
Mats-SX: I'd like to have Tobias' definition spelled out here as well, namely that _equal_ values are grouped together.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+  ** Relationships are <<incomparable>> to any value that is not also a relationship.
+- Paths
+  ** Paths are compared as if they were a list of alternating nodes and relationships of the path from the start node to the end node. For example, `[Node(1)-Rel(2)->Node(4)] \<= [Node(1)<-Rel(3)-Node(2)]` (assuming graph entities are sorted in ascending order by an integer id in a given implementation).
```
Mats-SX: I'd like that parenthesis to not have to be here, so I'll write up a slightly more self-contained example as an alternative:

For example, given nodes `n1`, `n2`, `n3`, and relationships `r1` and `r3`, and given that `n1 < n2 < n3` and `r1 < r3`, then the path `p1` from `n1` to `n3` via `r1` would be less than the path `p2` to `n1` from `n2` via `r3`. Expressed in terms of lists:

```
p1 < p2 
=> [n1-r1->n3] < [n1<-r3-n2] 
=> n1 < n1 || r1 < r3 || n3 < n2 
=> false || true || false 
=> true
```

Here, I don't see us being able to differentiate between paths of the same nodes and relationships, but different directions. Unless I'm wrong, it could be worth mentioning this.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+  ** Paths track the direction in which a relationship is traversed. Two paths that traverse the same nodes and relationships in the same order but differ in the direction in which they traverse a single relationship are not equal. The comparison order for paths is expected to align with this.
```
Mats-SX: As mentioned in the above comment, I don't see how the list approach is dealing with this.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+
+[[equality-def,equality]]
+==== Equality ====
+
+In order to align equality with <<comparability-def>>, we change equality of lists and maps that contain `null` values to treat those values in the same way as if they would have been compared outside of those lists and maps, as individual, simple values.
+
+Specifically, we propose to redefine how equality works for lists in Cypher.
+To determine if two lists `l1` and `l2` are equal, we propose two simple tests, as exemplified by the following:
+
+* `l1` and `l2` must have the same size, i.e. inversely `size(l1) <> size(l2>) \=> l1 <> l2`
```
Mats-SX: typo in `size(l2>)`

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+----
+[a1, a2, ...] = [b1, b2, ...]
+<=>
+a1 = b1 && a2 = b2 && ...
+----
+
+For clarity, we also repeat the *current* equality semantics of maps here. Under these current semantics, two maps `m1` and `m2` are considered equal if:
+
+* `m1` and `m2` have the same keys,
+** including keys that map to a `null` value (the order of keys as returned by `keys` does not matter here).
```
Mats-SX: if this refers to the function, rewrite it to use parentheses: `keys()`

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+
+* `l1` and `l2` must have the same size, i.e. inversely `size(l1) <> size(l2>) \=> l1 <> l2`
+* the pairwise elements of both `l1` and `l2` must be equal, i.e.
+----
+[a1, a2, ...] = [b1, b2, ...]
+<=>
+a1 = b1 && a2 = b2 && ...
+----
+
+For clarity, we also repeat the *current* equality semantics of maps here. Under these current semantics, two maps `m1` and `m2` are considered equal if:
```
Mats-SX: I propose to split this into its own subsection `maps` with two subsections within it `current` and `new`, containing the old and newly proposed behaviour in clearly separated sections.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+As a consequence of these changes, plain <<equality-def>> is not reflexive for all values (consider: `{a: null} = {a: null}`, `[null] = [null]`). However this was already the case (consider: `null = null` \=> `null`).
+
+Note that <<equality-def>> is reflexive for values that do not involve `null` though.
+
+[[orderability-equivalence]]
+=== Orderability and equivalence ===
+
+We propose that <<orderability-def>> and <<equivalence-def>> should be aligned with each other, i.e.
+
+`expr1` is equivalent to `expr2` if and only if they would be sorted before (or after resp.) any other non-equivalent value in the same way (i.e. they have the same position under orderability).
```
Mats-SX: Isn't easier to define it using the expression in parentheses alone? Like this:

`expr1` is equivalent to `expr2` if and only if they have the same position under orderability.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+
+Orderability and equivalence use <<missing-null,"missing" `null` values>>.
+
+[[orderability-def,orderability]]
+==== Orderability ====
+
+[[global-sort-order,global sort order]]
+We further propose that orderability be defined between any pair of values, where the result is always `true` or `false`, i.e. always defined.
```
Mats-SX: This sentence starts on its own; `further` here feels wrong.

Mats-SX: I suggest dropping the `, i.e. always defined`. Comparability is defined as well, even though it sometimes yields `null`.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+
+* `MAP` types
+** <<regular-map,Regular maps>>
+** `NODE`
+** `RELATIONSHIP`
+* `LIST OF ANY?`
+* `PATH`
+* `STRING`
+* `BOOLEAN`
+* `NUMBER` (`NaN` is treated as the largest number in orderability only, i.e. it is put after positive infinity)
```
Mats-SX: I suggest adding a subbullet replacing the parenthesis, saying:

`NaN` is ordered as the largest number.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+* `VOID` (i.e. the type of `null`)
+
+To give a concrete example, under this global sort order all Nodes are less than all Strings.
```
Mats-SX: No Capitals For Nodes Or Strings Here.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+
+Between values of the same type in the global sort order, orderability defers to comparability except that equality is overriden by equivalence as described below.
+
+The accompanying descending global sort order is the same order in reverse (i.e. it runs from `VOID` to `NODE`).
```
Mats-SX: `VOID` to `MAP`

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
Mats-SX: This definition will allow this equality test to evaluate to true: `[] <> `[null]`. Is this intentional?

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,309 @@
...
+An integer compared to a float
+[source, cypher]
+----
+RETURN 1 > 0.5 // should be true
+----
+
+A string compared to a boolean
+[source, cypher]
+----
+RETURN 'string' \<= true // should be null
```
Mats-SX: Incorrect backslash

Mats-SX:
This is a really nice CIP. You might want to close and create new PR after having considered mine, as this is getting hard to get an overview of.

boggle:
Thanks for the feedback, I think I want to first work on this PR, marking entries with `TADAA` whenever I feel I have addressed them; once done, I'll move to a new PR

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: No but this is why it gets introduced earlier, it's a bit of a smell but I'd rather add that to the type system at some point.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: That's needed to stop the asciidoc renderer from turning this into a unicode symbol (which looks off)

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: It can only happen for self-loops which is a weird but real corner case.

boggle: Oh wait, in that case, the direction doesn't matter. In all other cases, start and end node tell us from which side we came. I'll update.

boggle:
Re-opening due to overwhelmingly many and awesome feedback comments and rebasing. Thanks all!

PR #112, opened 2016-07-07
First draft of parameter syntax CIP

Mats-SX:


cip/CIP2016-07-07-Parameter-syntax.adoc
```diff
@@ -0,0 +1,91 @@
...
+:toc:
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Author:* Mats Rydberg <mats@neotechnology.com>
+
+[abstract]
+.Abstract
+--
+This CIP suggests changing syntax for parameter usage, from curly braces (`{}`) to dollars (`$`).
```
petraselmer: 'to dollars' is imprecise (it sounds as if this was talking about the currency)... I suggest: 
', from curly braces (`{}`) to dollars (`$`).' --> 'by replacing the curly braces notation (`{}`) with the dollar delimiter (`$`).'

cip/CIP2016-07-07-Parameter-syntax.adoc
```diff
@@ -0,0 +1,91 @@
...
+Curly brackets are used by the language for literal maps and property predicates, and this overloading may cause confusion in usage.
```
petraselmer: Please use either 'brackets' or 'braces' consistently, rather than swapping (earlier, 'braces' was used)

petraselmer: I suggest the following sentence instead (or something similar):
'As curly braces are already used by the language for literal maps.... predicates, their usage as parameter delimiters may lead to confusion when writing or reading queries.'

cip/CIP2016-07-07-Parameter-syntax.adoc
```diff
@@ -0,0 +1,91 @@
...
+Using a different, unique character causes parameters to be more visible for the human eye, and by using only a single character for the parameter instead of one opening and one closing, the syntax becomes somewhat cleaner.
```
petraselmer: I would suggest something along these lines: 
'In order to make parameters more readable, a single, unique and hitherto unused character'
'A single, unique and hitherto unused character used as a prefix increases the readability of parameters when compared with delimiting parameters using overloaded characters. Moreover, this allows for cleaner syntax.'

cip/CIP2016-07-07-Parameter-syntax.adoc
```diff
@@ -0,0 +1,91 @@
...
+
+
+== Motivation
+
+//Provide the rationale and justification for the proposed feature(s).
+
+The clause `MATCH (a {prop: {param}})` contains a nested curly brace structure, which is not ideal for human readability.
+The alternative `MATCH (a {prop: $param})` does a much better job at signalling the parameter usage to the reader.
+
+== Background
```
petraselmer: The sentences in this section are from the template, and can be removed?

cip/CIP2016-07-07-Parameter-syntax.adoc
```diff
@@ -0,0 +1,91 @@
...
+--
+
+toc::[]
+
+
+== Motivation
+
+//Provide the rationale and justification for the proposed feature(s).
+
+The clause `MATCH (a {prop: {param}})` contains a nested curly brace structure, which is not ideal for human readability.
```
petraselmer: I suggest 
'contains a nested curly brace structure, which is not ideal for human readability.' 
is reworded to something such as:
'uses a nested curly brace structure which is not easily readable.'

cip/CIP2016-07-07-Parameter-syntax.adoc
```diff
@@ -0,0 +1,91 @@
...
+The alternative `MATCH (a {prop: $param})` does a much better job at signalling the parameter usage to the reader.
```
petraselmer: I would suggest rewording
'does a much better job at signalling the parameter usage to the reader.' 
to something like:
'makes the parameter usage far more legible.'

cip/CIP2016-07-07-Parameter-syntax.adoc
```diff
@@ -0,0 +1,91 @@
...
+
+== Background
+
+This is further background regarding the CIP, should it be required or deemed useful.
+
+Please ensure that links to related documents (including other CIPs) are provided.
+
+== Proposal
+
+The concrete suggestion made by this CIP is to replace the parameter syntax with curly braces with a dollar sign.
```
petraselmer: This sentence is very confusing; please replace 
'The concrete suggestion made by this CIP is to replace the parameter syntax with curly braces with a dollar sign.'
with (or something similar):
'This CIP proposes using the dollar sign instead of curly braces to delimit parameters.'

cip/CIP2016-07-07-Parameter-syntax.adoc
Mats-SX: How about `far more visible`?

cip/CIP2016-07-07-Parameter-syntax.adoc
Mats-SX: I wrote it like this:
```
The concrete suggestion made by this CIP is to define the parameter syntax to use a dollar sign.
This conflicts with the current use of wrapping curly braces, which this CIP suggests removing.
```

cip/CIP2016-07-07-Parameter-syntax.adoc
```diff
@@ -0,0 +1,98 @@
...
+The alternative `MATCH (a {prop: $param})` does a much better job at signalling the parameter usage to the reader.
+Using a single leading character is a very common way of referencing variables in general programming languages.
+Specifically, the following list of languages use the dollar sign:
+
+- Scala (string interpolation)
+- PHP (variable definition)
+- AutoIt (variable definition)
+- Perl (scalar variables)
+- Ruby (global variables)
+- BASIC (trailing character for variables)
```
thobe: In BASIC the dollar sign is actually (afaik) used to specify that the type of the variable is String.

cip/CIP2016-07-07-Parameter-syntax.adoc
```diff
@@ -0,0 +1,98 @@
...
+[source, cypher]
+----
+WITH $`Der Parameter` AS p
+RETURN toString(p)
+----
+
+.Using a map parameter as a property predicate.
+[source, cypher]
+----
+MATCH (n $param)
```
thobe: It should be noted that the space is required here, since otherwise it will be taken as a single variable name `n$param`.

thobe:
It would be nice if this PR also updated the grammar.

Quite probably we should remove [`Sc`](http://unicode.org/cldr/utility/list-unicodeset.jsp?a=%5B%3ASc%3A%5D&abb=on&esc=on&g=&i=) from [`IdentifierStart`](https://github.com/opencypher/openCypher/blob/923cbc72746279be443d613905772172cb974d3b/grammar/basic-grammar.xml#L527) - i.e. not allow _any_ currency signs as the first character of an identifier.

Mats-SX:
CLG discussion comments have been addressed. This CIP should now be mergeable.

PR #114, opened 2016-07-11
Comparability cip

boggle:


cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
+= CIP2016-06-14 - Redefine Comparability and Equality, as well as Orderability and Equivalence
```
thobe: I would like it if we could frame these documents as part of the specification, and not just enhancement proposals that you look at once and then throw away. In that light this title is wrong, this document is not (/ should not be) about _re_-defining Comparability, Equality, Orderability, and Equivalence, it is about _the definition_ of those things.

The PR is about re-defining these things, but the document is about the (new) definition of them.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+We propose to redefine <<comparability-def>> and <<equality-def>>, as well as <<orderability-def>> and <<equivalence-def>> as detailed in this section.
+
+[[concepts]]
+=== Concepts
+
+Cypher today features four distinct concepts related to equality and ordering:
+
+[horizontal]
+[[comparability-con,comparability (concept)]]*Comparability*:: <<comparability-def,Comparability>> is used by the inequality operators (`>`, `<`, `>=`, `\<=`), and defines the underlying semantics of how to compare two values.
+[[equality-con,equality (concept)]]*Equality*:: <<equality-def,Equality>> is used by the equality operators (`=`, `<>`), the list membership operator (`IN`), and value joins, and defines the underlying semantics to determine if two values are the same in these contexts.
```
thobe: Do we really need to list "value joins" here? That is an implementation strategy thing, is it not? From the point of view of the language it is expressed as equality (`=`), and could have been implemented as a filter on a cartesian product.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+
+[horizontal]
+[[unknown-null,unknown `null`]]*Unknown*:: An "unknown" `null` is taken to be a placeholder for an arbitrary but unknown value. When evaluating predicates, an "unknown" `null` is the `maybe` truth value of ternary logic. For node and relationship properties, an "unknown" `null` is a value that is definite in the real world but has not been stored in the graph. Since in these cases, two "unknown" `null` values stand for arbitrary but definite values in the real world, two "unknown" `null` values should never be treated as certainly being the same value.
+[[missing-null,missing `null`]]*Missing*:: A "missing" `null` is taken to be a marker for the absence of a value. In the context of updating node properties from a map, a "missing" `null` is used to mark properties that are to be removed. In the context of `DISTINCT` and grouping, a "missing" `null` value is used as grouping key for all records that miss a more specific value. Since in these cases, two "missing" `null` values represent the same concept, they should always be treated as the same value.
+
```
thobe: I wonder how helpful it is to discuss different "meanings" of nulls, once the `null` is put in place of a value this "origin"/"meaning" is "lost", a `null` is a `null` at that stage, and the work the same, regardless or where it came from.

Or does this document also propose to introduce a distinction between these two different "types" of `null`?

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+Cypher today has one supertype `MAP` for all map values. This includes nodes (of subtype `NODE`), relationships (of subtype `RELATIONSHIP`), and any other map (not captured by a subtype of `MAP`). For the purpose of this document, we define a regular map to be any value of type `MAP` that is neither a `NODE` nor a `RELATIONSHIP`.
+
+[[comparability-equality,comparability and equality]]
+=== Comparability and equality
+
+We propose that comparability and equality should be aligned with each other, i.e.
+
+`expr1 = expr2` if and only if `expr1 >= expr2 && expr1 \<= expr2`.
+
+Comparability and equality use <<unknown-null,"unknown" `null` values>>.
```
thobe: `s/use/produce/`? Plus specify (link to) the conditions during which it produces such "values".

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+- Numbers
+  * Integers are compared numerically in ascending order.
+  * Floats (excluding `NaN` and the Infinities) are compared numerically in ascending order.
+  * Numbers of different types (excluding `NaN` and the Infinities) are compared to each other as if both numbers would have been coerced to arbitrary precision big decimals (currently outside the Cypher type system) before comparing them with each other numerically in ascending order.
+  * Positive infinity is of type `FLOAT`, equal to itself and larger than any other number (excluding `NaN`).
+  * Negative infinity is of type `FLOAT`, equal to itself and smaller than any other number (excluding `NaN`).
+  * `NaN` is <<incomparable>>.
+  * Numbers are <<incomparable>> to any value that is not also a number.
+- Booleans
+  * Booleans are compared such that `false` is less than `true`.
```
thobe: This seems arbitrary. Why do we need this?

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+  * Booleans are <<incomparable>> to any value that is not also a boolean.
+- Strings
+  * Strings are compared in dictionary order, i.e. characters are compared pairwise in ascending order from the start of the string to the end.
+  Characters missing in a shorter string are considered to be smaller than any other character.
```
thobe: I think we should (consistently) use "less than" rather than "smaller than".

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+  For example, `'a' < 'aa'`.
+  * Strings are <<incomparable>> to any value that is not also a string.
+- Lists
+  * Lists are compared in dictionary order, i.e. list elements are compared pairwise in ascending order from the start of the list to the end. Elements missing in a shorter list are considered to be smaller than any other value.
+  For example, `[1] < [1, 0]`.
+  * Lists containing `null` values are <<incomparable>>. For example, `[1] >= [1, null]` evaluates to `null`.
+  * Lists are <<incomparable>> to any value that is not also a list.
+- Maps
+  * [[regular-maps,regular maps]]Regular maps
+  ** The comparison order for maps is unspecified and left to implementations.
```
thobe: This seems strange. What is the benefit of specifying that these values _are_ comparable, but not specifying _how_ they compare? Why not simply let maps be incomparable?

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+  For exampe, `{a: 1} \<= {a: 1, b: null}` evaluates to `null`.
+  ** Regular maps are <<incomparable>> to any value that is not also a regular map.
+  * Nodes
+  ** The comparison order for nodes is unspecified and left to implementations.
+  ** Nodes are <<incomparable>> to any value that is not also a node.
+  * Relationships
+  ** The comparison order for relationships is unspecified and left to implementations.
+  ** Relationships are <<incomparable>> to any value that is not also a relationship.
+- Paths
+  ** Paths are compared as if they were a list of alternating nodes and relationships of the path from the start node to the end node.
```
thobe: Given that the comparability for nodes and relationships is *undefined*, it seems really counter intuitive to then _define_ comparability for paths based upon_ the undefined comparability of nodes and relationships.

Why do we need paths to be comparable at all?

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+  => n1 < n1 || r1 < r3 || n3 < n2
+  => false || true || false
+  => true
+
+  ** Paths are <<incomparable>> to any value that is not also a path.
+- Implementation-specific types
+  * Implementations may choose to define suitable comparability rules for values of additional, non-canonical types.
+  * Values of an additional, non-canonical type are expected to be <<incomparable>> to values of a canonical type.
+
+Any kind of value not covered here is considered <<incomparable>>. This includes comparing a `null` value to any other value.
```
thobe: This bullet point is superfluous, since all types already specify that they are incomparable to any other type...

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+* Additionally, for each such key `k`,
+** either `m1.k = m2.k` is `true`,
+** or both `m1.k IS NULL` and `m2.k IS NULL`
+
+This is at odds with the decision to use <<unknown-null,"unknown" `null` values>> in <<comparability-equality>>.
+
+However, this definition is aligned with the most common use case for maps with `null` entries: updating multiple properties through the use of a single `SET` clause, e.g. `SET n += { size: 12, remove_this_key: null }`. In this case, there is no need to differentiate between different `null` values, as `null` merely serves as a marker for keys to be removed (i.e. is a <<missing-null,"missing" `null` value>>). Current equality semantics make it easy to check if two maps would correspond to the same property update in this scenario. We note though that this type of update map comparison is rare and could be emulated using a more complex predicate. The current rules do however break symmetry with how equality handles `null` in all other cases. This becomes more apparent by considering these two examples:
+
+* `expr1 = expr2` evaluates to `null` if `expr1 IS null && expr2 IS NULL`
```
thobe: inconsistent casing, `null` and `NULL`.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+Note that <<equality-def>> is reflexive for values that do not involve `null` though.
+
+[[orderability-equivalence]]
+=== Orderability and equivalence ===
+
+We propose that <<orderability-def>> and <<equivalence-def>> should be aligned with each other, i.e.
+
+`expr1` is equivalent to `expr2` if and only if they have the same position under orderability (i.e. they would be sorted before (or after resp.) any other non-equivalent value in the same way).
+
+Orderability and equivalence use <<missing-null,"missing" `null` values>>.
```
thobe: Again, I feel the word "use" is too vague, and would like to see a more explicit description of what is actually meant.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+* `MAP` types
+** <<regular-map,Regular map>>
+** `NODE`
+** `RELATIONSHIP`
+* `LIST OF ANY?`
+* `PATH`
+* `STRING`
+* `BOOLEAN`
+* `NUMBER`
+** `NaN` is treated as the largest number in orderability only (i.e. it is put after positive infinity)
```
thobe: `s/largest/greatest/`, we should (consistently) talk about values being _less than_ or _greater than_. We also shouldn't talk about it being "put after" without specifying whether the order is ascending or descending.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+[[equivalence-def,equivalence]]
+==== Equivalence ====
+
+Equivalence remains mostly unchanged but now can be defined succinctly as being identical to equality except that:
+
+* Any two `null` values are equivalent (both directly or inside nested structures).
+* Any two `NaN` values are equivalent (both directly or inside nested structures).
+* However, `null` and `NaN` values are not equivalent (both directly or inside nested structures).
+
+<<equivalence-def,Equivalence>> is reflexive for all values.
```
thobe: This links to the section it is in, that is a confusing link.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+=== Aggregation ===
+
+Generally, an aggregation function call `aggr(expr)` processes all matching rows for each aggregation key found in an incoming record (keys are compared using equivalence, as noted above).
+For a fixed aggregation key and each matching record, `expr` is evaluated to a value.
+This yields a list of candidate values.
+
+In a regular aggregation (i.e. of the form `aggr(expr)`), the list of aggregated values is the list of candidate values with all `null` values removed from it.
+
+In a distinct aggregation (i.e. of the form `aggr(DISTINCT expr)`), the list of aggregated values is the list of candidate values with all `null` values removed from it.
+Futhermore, in a distinct aggregation, only the first of all equivalent candidate values is included in the list of aggregated values, i.e. duplicates under equivalence are removed.
```
thobe: I am not convinced it is helpful to say that "only the first of all equivalent candidate values is included", since there is not necessarily an order defined for the aggregation. Could it be sufficient to say that "only one of the equivalent candidate values is included"

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,367 @@
...
+----
+RETURN 'string' <= true // should be null
+----
+
+Ordering values of different types
+[source, cypher]
+----
+UNWIND [1, true, '', 3.14, {}, [2]] AS i
+RETURN i
+  ORDER BY i // should not crash
```
thobe: We can do better than "should not crash", given the orderability definition, we should be able to tell the exact order in which these items are returned.

thobe:
I would propose that instead of saying that some types are comparable, but not defining how they are comparable, we should specify that they are *not* comparable, and instead specifically define their orderability.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
Mats-SX: The correct casing for `IS NULL` and `IS NOT NULL` is with capitals.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: This is introduced because we refer to these notions in the text, also I hope it helps the reader to understand the subtle differences between equality and equivalence. Distinguishing between different kinds of `null` is discussed in alternatives but not proposed here, and the case is made there, that even given that, we'd still need equality and equivalence as separate concepts.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: This decision is not completely arbitrary, even if it may appear so: Most languages (including Java and SQL) sort that way. I also think that there is an underlying algebraic reason for that, `false` is the neutral element of conjunction (which can be seen as logical addition, i.e. `false` corresponds to a `0`). `true` is the neutral element of disjunction (which can be seen as logical multiplication, i.e. `true corresponds to a `1`). `0 < 1` and by extension `false < true`

boggle: The decision to make booleans comparable at all comes from three points:

- Common languages define that
- The goal of this CIP is to make all properties orderable. This enables a couple of nice things, like `MATCH (n) RETURN n ORDER BY n.prop` will produce a predictable result and never fail.
- To keep a simple model, orderability is defined in terms of comparability.
- The cost of defining something that is not used by a given user is small.

These points together added up to that decision.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: The precise way in which nodes and relationships is left to the implementation however an implementation is required to define an order for them (I now made this more clear in the text). Paths simply piggyback on that.

Doing so, gives a nice lexicographic order on paths, i.e. you see them in some traversal like order when sorting them. I think that is useful.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: I wrote it as a catch-all, but agreed, let's remove it.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: The idea here was to use different wording to distinguish between orderability and comparability. Do you think that works if carried out consistently thoughout the text?

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: At least today this is the semantics, so the values are aggregated in the same order as the incoming records. This is relied upon sometimes in queries that use `collect`. Do you feel we should change that? We could differentiate depending on if this happens in an `ORDER BY` context or not.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: @systay felt that this should be left to implementations. The original idea was to sort by keys and then sort in dictionary order. Let's discuss comparability vs orderability at the bottom (your current last comment).

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
thobe: I agree that we want this for _orderability_, but I maintain my position that values could be orderable without being comparable. I.e. you don't necessarily have to be able to apply the `<` and `>` operators to the values in order to be able to `ORDER BY` them. I think it would be sensible for orderability to be "broader" than comparability, as long as it agrees with comparability in all cases where comparability applies.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
thobe: I do.

cip/CIP2016-06-14-Redefine-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
thobe: I feel like _order of incoming records_ is not something we want to specify, and thus should not be something we should reference. Imagine we'd want to parallelise evaluation in the future - that makes it hard to specify such a "first" candidate.

It seems sufficient to just say that we will pick _one_.

boggle:
So the thinking here is that making them comparable causes no harm, after all this could only become relevant if someone compared two values that they'd expect to have a different type. 
However aligning comparability and orderability gives a really simple mental model, and I think that's more valuable than having slightly arbitrary comparison rules between values no one usually compares.

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,371 @@
...
+- Maps
+  * [[regular-maps,regular maps]]Regular maps
+  ** The comparison order for maps is unspecified and left to implementations.
+  ** The comparison order for maps must align with the <<equality-def,equality semantics>> outlined below.
+  In consequence, any map that contains an entry that maps its key to a `null` value is <<incomparable>>.
+  For exampe, `{a: 1} \<= {a: 1, b: null}` evaluates to `null`.
+  ** Regular maps are <<incomparable>> to any value that is not also a regular map.
+  * Nodes
+  ** The comparison order for nodes is unspecified here and left to implementations.
+  However implementations are required to define a fixed comparison oder between nodes.
```
thobe: s/oder/order/

boggle:
Note: s/self.?loop/loop/

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,377 @@
...
+  ** Relationships are <<incomparable>> to any value that is not also a relationship.
+- Paths
+  ** Paths are compared as if they were a list of alternating nodes and relationships of the path from the start node to the end node.
+  For example, given nodes `n1`, `n2`, `n3`, and relationships `r1` and `r3`, and given that `n1 < n2 < n3` and `r1 < r3`, then the path `p1` from `n1` to `n3` via `r1` would be less than the path `p2` to `n1` from `n2` via `r3`. Expressed in terms of lists:
+
+  p1 < p2
+  => [n1-r1->n3] < [n1<-r3-n2]
+  => n1 < n1 || r1 < r3 || n3 < n2
+  => false || true || false
+  => true
```
thobe: this example does not illustrate the actual comparison, since it deals only with `true` and `false` and _or_. Actual ordering comparison would have to deal with _less than_, _greater than_, and _equal_, and combining that.

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,377 @@
...
+
+  ** Additionally, paths track the direction in which a relationship is traversed.
+  Two paths that traverse the same nodes and relationships in the same order but differ in the direction in which they traverse a single relationship are not equal (this can happen in paths containing self-loops).
+  The comparison order for paths is expected to align with this.
```
thobe: We should specify exactly _how_ direction should be taken into account for ordering.

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,377 @@
...
+* Any two `null` values are equivalent (both directly or inside nested structures).
+* Any two `NaN` values are equivalent (both directly or inside nested structures).
+* However, `null` and `NaN` values are not equivalent (both directly or inside nested structures).
+
+Equivalence is reflexive for all values.
+
+[[aggregation]]
+=== Aggregation ===
+
+Generally, an aggregation function call `aggr(expr)` processes all matching rows for each aggregation key found in an incoming record (keys are compared using equivalence, as noted above).
```
thobe: I don't think of aggregation as "function calls", could we not just call it "aggregation", rather than "aggregation function call"?

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -0,0 +1,377 @@
...
+For a fixed aggregation key and each matching record, `expr` is evaluated to a value.
+This yields a list of candidate values.
+Generally, the order of candidate values is unspecified.
+However, if the aggregation happens in a projection with an associated `ORDER BY` subclause, the list of candidate values is ordered in the same way as the underlying records and as specified by the associated `ORDER BY` subclause.
+
+In a regular aggregation (i.e. of the form `aggr(expr)`), the list of aggregated values is the list of candidate values with all `null` values removed from it.
+
+In a distinct aggregation (i.e. of the form `aggr(DISTINCT expr)`), the list of aggregated values is the list of candidate values with all `null` values removed from it.
+Futhermore, in a distinct aggregation, only the first of all equivalent candidate values is included in the list of aggregated values, i.e. duplicates under equivalence are removed.
```
thobe: We should not say "only the first of all equivalent", we should say "only one of all equivalent", in order to not make any specific promises or requirements about _which_ value is picked. It _shouldn't_ matter, since they are equivalent, but let's not promise more than we need to in case it _could_ matter.

thobe: We could make an additional note saying that when used with `ORDERD BY` the _first of the equivalent as defined by the ordering_ is picked. But that should be an additional note, we should still say that _only one of all equivalent values_ is picked in the general case.

boggle:
Note: Either `l <= r || l >= r`

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: Let's take that on the next discussion round; for now left to implementations.

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
boggle: But `count` etc are commonly called aggregation functions? Happy to change though.

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -138,14 +139,17 @@ We propose that <<comparability-con,comparability>> should be defined between an
...
-  => false || true || false
-  => true
+      p1 < p2
+  <=> [n1-r1->n3] < [n1<-r3-n2] ?
+  <=> n1 < n1 || (n1 = n1 && [r1, n3] < [r3, n2])
+  <=> false || (true && [r1, n3] < [r3, n2])
+  <=> [r1, n3] < [r3, n2]
+  <=> r1 < r2 || (r1 = r3 && n3 < n2)
+  <=> true || (false && false)
+  <=> true
```
Mats-SX: These should not be equivalences, but implications.

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -138,14 +139,17 @@ We propose that <<comparability-con,comparability>> should be defined between an
   ** Paths are compared as if they were a list of alternating nodes and relationships of the path from the start node to the end node.
   For example, given nodes `n1`, `n2`, `n3`, and relationships `r1` and `r3`, and given that `n1 < n2 < n3` and `r1 < r3`, then the path `p1` from `n1` to `n3` via `r1` would be less than the path `p2` to `n1` from `n2` via `r3`. Expressed in terms of lists:
 
-  p1 < p2
-  => [n1-r1->n3] < [n1<-r3-n2]
-  => n1 < n1 || r1 < r3 || n3 < n2
-  => false || true || false
-  => true
+      p1 < p2
+  <=> [n1-r1->n3] < [n1<-r3-n2] ?
```
Mats-SX: stray question mark

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -138,14 +139,17 @@ We propose that <<comparability-con,comparability>> should be defined between an
...
+  <=> n1 < n1 || (n1 = n1 && [r1, n3] < [r3, n2])
+  <=> false || (true && [r1, n3] < [r3, n2])
+  <=> [r1, n3] < [r3, n2]
+  <=> r1 < r2 || (r1 = r3 && n3 < n2)
```
Mats-SX: `r3`, not `r2`.

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -228,39 +233,46 @@ To accomplish this, there must be a pre-determined order of types and each value
...
+Generally the order of candidate values is unspecified.
+If the aggregation happens in a projection with an associated `ORDER BY` subclause, the list of candidate values is ordered in the same way as the underlying records and as specified by the associated `ORDER BY` subclause.
 
 In a regular aggregation (i.e. of the form `aggr(expr)`), the list of aggregated values is the list of candidate values with all `null` values removed from it.
 
 In a distinct aggregation (i.e. of the form `aggr(DISTINCT expr)`), the list of aggregated values is the list of candidate values with all `null` values removed from it.
-Futhermore, in a distinct aggregation, only the first of all equivalent candidate values is included in the list of aggregated values, i.e. duplicates under equivalence are removed.
+Futhermore, in a distinct aggregation, only one of all equivalent candidate values is included in the list of aggregated values, i.e. duplicates under equivalence are removed.
```
Mats-SX: Fu`r`thermore

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -228,39 +233,46 @@ To accomplish this, there must be a pre-determined order of types and each value
...
+However, if the distinct aggregation happens in a projection with an associated `ORDER BY` subclause, always the first from a set of equivalent candidate values is included in the list of aggregated values.
```
Mats-SX: Shouldn't it be `only one element from each set of equivalent candidate values ...`?

cip/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -346,18 +360,20 @@ This CIP introduces the distinction between "unknown" and "missing" `null` value
 
 A consistent set of rules is defined for <<equality-con,equality>>, <<equivalence-con,equivalence>>, <<comparability-con,comparability>> and <<orderability-con,orderability>>.
 
+Futhermore, aggregation semantics are clarified and this proposal prepares the replacement (or reinterpretation) of `NaN` values as `null` values in the future.
```
Mats-SX: Fu`r`thermore.

Mats-SX:
@boggle I opted for a simple sentence instead of the appendix with `cmp` and `ord`. I didn't find what I managed to write on them useful.

Mats-SX:
Needs rebasing to end up in the correct directory (`accepted`).

PR #120, opened 2016-07-21
EXISTS CIP

Mats-SX:
Supersedes #10.

### DIFF
--- a/cip/1.accepted/CIP2015-05-13-EXISTS.adoc
+++ b/cip/1.accepted/CIP2015-05-13-EXISTS.adoc
@@ -22,7 +22,7 @@ This form is helpful, but not exhaustive in the types of subqueries that users w
 
 == Proposal
 
-To make this feature more powerful, add a new EXISTS-function, that would exist in two forms:
+To make this feature more powerful, add a new function named +EXISTS+, that would exist in two forms:
 
  * Property existence checking
  * Subquery existence checking
@@ -51,15 +51,14 @@ Note: the read-only clause above includes WITH but not RETURN.
 === Semantics
 
 All forms of +EXIST(...)+ accomplish the same task - check if a particular pattern exists in the database.
-They are expressions, and as such must be side-effect free. That is, the subqueries in +EXIST+ must not be updating queries.
+They are expressions, and as such must be side-effect free; that is, the subqueries in +EXIST+ must not be updating queries.
 Regarding nomenclature, all forms of +EXISTS+ are scalar expressions - they return a single boolean value, and this value is never +null+.
 
 ==== Property exists
 
 In the property checking form, any expression is accepted as the input.
-If the input expression returns a +null+ value, +EXIST(...)+ returns false, and for all other values, it will return +true+.
-This form of +EXISTS(...)+ is equivalent to checking for +null+ using +IS NOT NULL+.
-I.e. +EXISTS(expr)+ is equivalent to +expr IS NOT NULL+.
+If the input expression returns a +null+ value, +EXISTS(...)+ returns false, and for all other values, it will return +true+.
+This form of +EXISTS(...)+ is equivalent to checking for +null+ using +IS NOT NULL+; i.e. +EXISTS(expr)+ is equivalent to +expr IS NOT NULL+.
 
 This form is there to make property existence checking idiomatic in Cypher - we do not want the user thinking of missing properties as checking for +null+ values.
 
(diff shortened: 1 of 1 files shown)


cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,175 @@
...
+Variables introduced in the subquery are not available on the outside.
+
+For each matching subgraph evaluated with `EXISTS {}`, the result value must be `true` if the subquery finds at least one matching row.
+If no matches are found, `false` should be returned.
+
+The subquery used in `EXISTS {}` has to follow the following rules:
+
+ * It has to be a read-only query. Updating the graph as part of a predicate is not allowed.
+ * It cannot end in a `RETURN` clause, nor may it contain `UNION`.
+ * If `WITH` ends the subquery, it should be able to change the cardinality of the subquery.
```
alastai: This is test comment by Alastair

alastai: This clause is redundant and contradicts the grammar definition of 

> read only clause

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,175 @@
...
+In the property checking form, any expression is accepted as the input.
+If the input expression evaluates to `null`, `exists()` will evaluate to `false`. For any other value, it will evaluate to `true`.
+This form of `exists()` is equivalent to checking for `null` using `IS NOT NULL`: `exists(expr)` is equivalent to `expr IS NOT NULL`.
+
+This form is there to make property existence checking idiomatic in Cypher - we do not want the user thinking of missing properties as checking for `null` values.
+
+==== Subquery exists
+
+When checking for the existence of subqueries, variables and parameters from the outer query are available in the subquery.
+Variables introduced in the subquery are not available on the outside.
```
petraselmer: Perhaps say something about this being a correlated subquery? We've done the same for the 'general' subqueries CIPed elsewhere...

cip/CIP2015-05-13-EXISTS.adoc
```diff
@@ -0,0 +1,175 @@
...
+For each matching subgraph evaluated with `EXISTS {}`, the result value must be `true` if the subquery finds at least one matching row.
+If no matches are found, `false` should be returned.
+
+The subquery used in `EXISTS {}` has to follow the following rules:
+
+ * It has to be a read-only query. Updating the graph as part of a predicate is not allowed.
+ * It cannot end in a `RETURN` clause, nor may it contain `UNION`.
+ * If `WITH` ends the subquery, it should be able to change the cardinality of the subquery.
+ In other words, the `WITH` clause may include `WHERE` or `SKIP`, since these sub-clauses can turn a query that returns rows into one that does not.
+
```
petraselmer: Is there a reason why EXISTS {...} cannot be nested? At the very least, this ought to be mentioned.

cip/CIP2015-05-13-EXISTS.adoc
Mats-SX: I'm not sure I understand what a 'correlated subquery' is. Do you have a reference?

cip/CIP2015-05-13-EXISTS.adoc
Mats-SX: I think that you might've misread this bullet point. It does state that `WITH` is allowed, and that is is further allowed to also change the cardinality of the subquery, through using `SKIP`, `LIMIT`, and most importantly `WHERE`. By changing cardinality like that, it does have the potential to alter the result of the subquery.

cip/CIP2015-05-13-EXISTS.adoc
Mats-SX: I don't think there is such a reason. I'll add a statement that makes it explicitly possible.

cip/CIP2015-05-13-EXISTS.adoc
Mats-SX: Perhaps the text is unintentionally easy to misunderstand? Perhaps we need to reword this?

Mats-SX: Suggestion: 

If `WITH` ends the subquery, it is able to change the cardinality.

alastai: You are right. If we change the words "should be able" to "is able" that would be clearer.

Issue #131, opened 2016-08-13
Grammar prohibits using whitespace between pattern parts in MATCH clause


jmarton:
The current grammar (at least the generated ANTLR4 grammar) only allows the 1st form, and rejects the 2nd and 3rd query forms below. I.e. it prohibits using whitespaces around the comma between the `PatternPart` non-terminal symbols.

1. `MATCH (n1),(n2) RETURN n1, n2`
2. `MATCH (n1), (n2) RETURN n1, n2`
3. `MATCH (n1) , (n2) RETURN n1, n2`

PR #132, opened 2016-08-13
Allow whitespace between PatternPart symbols of a MATCH clause.

jmarton:
Fixes opencypher/openCypher#131

### DIFF
--- a/grammar/basic-grammar.xml
+++ b/grammar/basic-grammar.xml
@@ -39,7 +39,7 @@
 
   <production name="Pattern">
     <non-terminal ref="PatternPart"/>
-    <repeat> , <non-terminal ref="PatternPart"/></repeat>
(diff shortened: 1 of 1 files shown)


Mats-SX:
@jmarton Looks great, thanks for your contribution.

The current testing procedure that we use is to add known correct queries to this file: [cypher.txt](https://github.com/opencypher/openCypher/blob/master/tools/grammar/src/test/resources/cypher.txt) (and [cypher-legacy.txt](https://github.com/opencypher/openCypher/blob/master/tools/grammar/src/test/resources/cypher-legacy.txt) for queries that use legacy constructs). These will then be parsed through the Antlr grammar in a test run by CI, to verify that the fix worked (and that nothing else broke).

I'll add this separately in a commit after merging yours (as I don't have write access to your fork).

PR #153, opened 2016-10-20
[Grammar] Add missing Expression in filter expression

FylmTM:
Hi,

I have noticed that `Expression` is missing after `WHERE` in `FilterExpression`.

So, here it is.

### DIFF
--- a/grammar/basic-grammar.xml
+++ b/grammar/basic-grammar.xml
@@ -301,7 +301,7 @@
 
   <production name="FilterExpression">
     <non-terminal ref="IdInColl"/>
-    <opt>&WS;<non-terminal ref="Where"/></opt>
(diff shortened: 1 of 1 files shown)


Mats-SX:
Hi @FylmTM and apologies for the late reply.

The right-hand side expression is baked into the `Where` rule:

```
  <production name="Where" rr:inline="true">
    WHERE &SP; <non-terminal ref='Expression'/>
  </production>
```
(I've expanded the `&expr;` macro above for clarity.)

Is there a particular query that is failing to parse for you?

FylmTM:
Hi @Mats-SX

Sorry. It was my fault. I had error in grammar rule in plugin in this particular case.

PR #154, opened 2016-11-02
Add definition of eagerness

Mats-SX:


petraselmer:
(In the abstract)
"This CIP tries to formalise how state changes behave while inside a Cypher query.
 Currently, because this is not explicit enough, the behaviour is implementation dependent."
->
"This CIP formalises the behaviour of state changes within a Cypher query."

cip/1.accepted/CIP2015-10-27-State-visibility-between-clauses.adoc
```diff
@@ -16,27 +16,35 @@ Currently, because this is not explicit enough, the behaviour is implementation
...
 Some clauses can also do both - read the graph and write to it at the same time.
 
 To allow the user to understand what the outcome will be at an abstract, logical level, semantics for the visibility of these changes should be clear.
 
 Note:
 This is not about concurrency - even in a single user system this would need to be thought through and defined explicitly.
 
+===== Definition
+
+We call the concept discussed in this CIP _Eagerness_.
```
petraselmer: (a suggestion)
" _Eagerness_ " -> " '_eagerness_' "

cip/1.accepted/CIP2015-10-27-State-visibility-between-clauses.adoc
```diff
@@ -16,27 +16,35 @@ Currently, because this is not explicit enough, the behaviour is implementation
...
+The meaning here is that each clause is logically consuming _all_ incoming rows _eagerly_, without producing any output until the input is exhausted.
```
petraselmer: "This means that each clause logically consumes _all_ incoming rows _eagerly_, and only produces output when the input is exhausted."

cip/1.accepted/CIP2015-10-27-State-visibility-between-clauses.adoc
```diff
@@ -16,27 +16,35 @@ Currently, because this is not explicit enough, the behaviour is implementation
...
+This concept is the opposite of link:https://en.wikipedia.org/wiki/Lazy_evaluation[_lazy evaluation_], which applied to Cypher means that each clause only processes as much input as is necessary for the first record of output to be sent to the subsequent clause.
```
petraselmer: "This concept is the opposite of" -> "This concept is in contrast to"

"which applied to Cypher means that each clause only processes as much input as is necessary for the first record of output to be sent to the subsequent clause."
->
"which, when applied to Cypher, means that each clause only processes as much input as is necessary for the first record of output to be sent to the subsequent clause."

Mats-SX:
@petraselmer Is this now ready for merge?

PR #155, opened 2016-11-02
Add styling rules for subquery `EXISTS`

Mats-SX:
Restructure content into sections, based on type of formatting (rather than type of language construct).

### DIFF
--- a/docs/style-guide.adoc
+++ b/docs/style-guide.adoc
@@ -345,23 +345,6 @@ MATCH (:Person)-->(vehicle:Car)-->(:Company)<--
 RETURN count(vehicle)
 ----
 
-. Surround node patterns with parentheses.
-.. This only concerns old Cypher code, as omitting parentheses is no longer legal in Cypher.
-+
-.Bad
-[source, cypher]
-----
-MATCH person-->vehicle
-RETURN person, vehicle
-----
-+
-.Good
-[source, cypher]
-----
-MATCH (person)-->(vehicle)
-RETURN person, vehicle
-----
-
 . Use anonymous nodes and relationships when the variable would not be used.
 +
 .Bad
--- a/docs/style-guide.adoc
+++ b/docs/style-guide.adoc
@@ -107,6 +107,58 @@ RETURN a.prop
 .. Put `ON CREATE` before `ON MATCH` if both are present.
 // no example necessary
 
+. Start a subquery on a new line after the opening brace, indented with two (additional) spaces.
++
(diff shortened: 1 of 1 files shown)


docs/style-guide.adoc
```diff
@@ -237,73 +287,81 @@ WITH null AS n1, null as n2
...
 ----
-MATCH (node:NonSpacedLabel {property: 42})
-RETURN labels(node)
+MATCH (:Person)-->(vehicle:Car)-->(:Company)<--
+      (:Country)
+RETURN count(vehicle)
 ----
 
-. Do not use a semicolon at the end of the statement.
+. Surround node patterns with parentheses.
```
petraselmer: I wonder whether we should zap this section (surrounding node patterns with parentheses) entirely? I don't think there is much value about talking about old Cypher code here...

docs/style-guide.adoc
```diff
@@ -415,6 +473,7 @@ RETURN map
 ----
 
 . No padding space for parameters.
+.. This rule mentions deprecated syntax. See link:https://github.com/opencypher/openCypher/blob/master/cip/2.testable/CIP2016-07-07-Parameter-syntax.adoc#interaction-with-existing-features[Parameter Syntax].
```
petraselmer: In contrast to my comment about surrounding nodes with parentheses, this is one rule I would definitely keep for the time being 👍

docs/style-guide.adoc
```diff
@@ -526,6 +585,24 @@ RETURN split( 'original', 'i' )
 RETURN split('original', 'i')
 ----
 
+. Use padding space on the inside of simple subquery expressions.
```
petraselmer: "Use padding space on the inside of simple subquery expressions."
->
"Use padding space within simple subquery expressions."

docs/style-guide.adoc
```diff
@@ -105,120 +107,168 @@ RETURN a.prop
...
+----
++
+.Good
+[source, cypher]
+----
+MATCH (e:EditorInChief)-->(:Employee)
+RETURN e.name
+----
+
+. Write relationship types in upper case, using underscore to separate words.
```
Mats-SX: How about 

> an underscore (`_`)

docs/style-guide.adoc
Mats-SX: I have been thinking the same thing on several occasions. Now you provided me the courage necessary. Will zap.

PR #158, opened 2016-11-03
First draft of specifying the Property Graph Model

Mats-SX:


docs/property-graph-model.adoc
```diff
@@ -0,0 +1,75 @@
+:numbered:
+
+= Property Graph Model
+ifdef::env-github,env-browser[:outfilesuffix: .adoc]
+
+Cypher is a graph query language, but it doesn't operate on any arbitrary definition of graph; it operates on a Property Graph.
```
petraselmer: "Cypher is a graph query language, but it doesn't operate on any arbitrary definition of graph; it operates on a Property Graph."
->
"Cypher is a graph query language which operates on a _property graph_."

docs/property-graph-model.adoc
```diff
@@ -0,0 +1,75 @@
...
+A Property Graph may be defined in graph theoretical terms as a directed, vertex-labeled, edge-labeled multigraph with self-edges, where edges have their own identity.
```
petraselmer: "Property Graph"
->
"property graph"

petraselmer: Please apply this change elsewhere in the document

petraselmer: Also mention that we use the term "node" to denote a vertex, and "relationship" to denote an edge (since "vertex" and "edge" are used to define the property graph model)

docs/property-graph-model.adoc
```diff
@@ -0,0 +1,75 @@
...
+* Path
+* Token
+** Label
+** Relationship type
+** Property key
+* Property
+
+=== Entity
+
+- An entity has a unique, comparable identity, which defines whether two entities are equal or not.
```
petraselmer: "comparable identity, which defines whether two entities are equal or not"
->
"comparable identity which defines whether or not two entities are equal"

docs/property-graph-model.adoc
```diff
@@ -0,0 +1,75 @@
...
+- An entity is assigned a set of properties, which are each uniquely identified in the set by their respective property keys.
```
petraselmer: "which are each uniquely identified"
->
"each of which are uniquely identified"

docs/property-graph-model.adoc
```diff
@@ -0,0 +1,75 @@
...
+
+==== Node
+
+- A _node_ is the basic entity of the graph, with the unique attribute of being able to exist in and of itself.
+- A node is assigned a set of unique labels.
```
petraselmer: Also mention that a node need not be labelled

docs/property-graph-model.adoc
```diff
@@ -0,0 +1,75 @@
...
+- A node may have zero or more outgoing relationships.
+- A node may have zero or more incoming relationships.
+
+==== Relationship
+
+- A _relationship_ is an entity that encodes a directed connection between exactly two nodes, the _start node_ and the _end node_.
+- A relationship's direction is said to be _outgoing_ from the start node, and _incoming_ to the end node.
```
petraselmer: Maybe replace with:
"An outgoing relationship is a directed relationship from the point of view of its source node."
and
"An incoming relationship is a directed relationship from the point of view of its target node."

docs/property-graph-model.adoc
```diff
@@ -0,0 +1,75 @@
...
+==== Node
+
+- A _node_ is the basic entity of the graph, with the unique attribute of being able to exist in and of itself.
+- A node is assigned a set of unique labels.
+- A node may have zero or more outgoing relationships.
+- A node may have zero or more incoming relationships.
+
+==== Relationship
+
+- A _relationship_ is an entity that encodes a directed connection between exactly two nodes, the _start node_ and the _end node_.
```
petraselmer: I think _source_ and _target_ node are better terms to use and are more in line with formal definitions (as opposed to 'start' and 'end', which sound a bit amateurish)

docs/property-graph-model.adoc
```diff
@@ -0,0 +1,75 @@
...
+- A relationship's direction is said to be _outgoing_ from the start node, and _incoming_ to the end node.
+- A relationship is assigned exactly one relationship type.
+
+=== Path
+
+- A _path_ is an odd-sized, ordered sequence of alternating nodes and relationships.
```
petraselmer: The first part of this sentence (odd-sized???) reads awkwardly.
Perhaps replace with all or part of the following sentence:
"A path represents a walk through a property graph and consists of a sequence of alternating nodes and relationships, starting and ending with a node."

docs/property-graph-model.adoc
Mats-SX: I thought that was clear as a set may be empty. Do you think this needs additional mention? Would the following be better (which we discussed):

> A node may be assigned a set of unique labels.

docs/property-graph-model.adoc
Mats-SX: I was trying to choose words here that align with other documentation around Cypher, such as the functions `startNode()` and `endNode()`. Do you think we should regard this as a free-standing source of truth and do the Right Thing here and change other places instead?

docs/property-graph-model.adoc
Mats-SX: Hmm, this almost makes it sound like there are two kinds of directed relationships, and possibly undirected ones. Is there something in particular about the sentence you dislike? Is it mainly making it two separate points (one for incoming and one for outgoing)?

docs/property-graph-model.adoc
Mats-SX: Do we then need to refer or define what a walk is, or is that well understood do you think?

docs/property-graph-model.adoc
petraselmer: For example, please see page 9 of https://arxiv.org/pdf/1409.3288v2.pdf (reconciling property graphs and RDF)

docs/property-graph-model.adoc
petraselmer: It is mainly about the latter point. Line 39 already specifies that we consider digraphs only, so I don't think an assumption will be made about our model allowing undirected edges.

docs/property-graph-model.adoc
petraselmer: Hmmm. I think we ought to define things from first principles - which would perhaps mean defining a walk too. I think, though, that this needn't come in just yet (that is to say, in this PR).

docs/property-graph-model.adoc
Mats-SX: I'm fine with viewing this document like so.

On another note: the paper you linked writes Property Graph With Capitals -- but I take it they are right in some regards but not all.

docs/property-graph-model.adoc
Mats-SX: I'm not 100% convinced, but I can't come up with a better alternative, so I'll use your formulations.

docs/property-graph-model.adoc
Mats-SX: I agree. I changed it, but I kept the `start and end with node` bit in a separate point. Is that OK?

petraselmer: That's ok

PR #159, opened 2016-11-08
Minor suspected typo, and end of list

fickludd:


cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -118,7 +118,7 @@ We propose that <<comparability-con,comparability>> should be defined between an
   * Strings are <<incomparable>> to any value that is not also a string.
 - Lists
   * Lists are compared in dictionary order, i.e. list elements are compared pairwise in ascending order from the start of the list to the end. Elements missing in a shorter list are considered to be less than any other value (including `null` values).
-  For example, `[1] < [1, 0]` but also `[1] \<= [1, null]`.
+  For example, `[1] < [1, 0]` but also `[1] \< [1, null]`.
```
Mats-SX: If you're removing the equals, you'll need to remove the escaping backslash too (it was a rendering hint, but only applies to `<=`, not `<`: see https://github.com/fickludd/openCypher/blob/850ae2214d5a1d759c0dffe763277688b077134a/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc#322-comparability).

I think the equals was there intentionally, but perhaps it is a source of confusion if we're showcasing two examples in one like that?

fickludd: Aha, well the issue I'm having is that I think the example clashes with the sentence before, where absent elements are claimed to be less that `null`. Then `[1] <> [1,null]`, which should also match the notion of the unknown null, meaning that the lists cannot be identical.

PR #160, opened 2016-11-18
Change definition of path comparability

Mats-SX:
The direction in which a relationship is traversed is
not significant for a path. The mention of this in the CIP
was caused by a misunderstanding.


thobe:
How can the direction not be important? If the direction is not important that would make `(a)-[x]->(a)` the same as `(a)<-[x]-(a)`, which I would argue that it is not.

Mats-SX:
Replaced by #162

PR #162, opened 2016-12-06
Add scenarios around pattern matching and self-loops

Mats-SX:
There is lots of confusion around how self-relationships (or loops) are
matched by patterns. These scenarios aim to outline how this is supposed
to work, and how it differs from simple relationship matching.

### DIFF
--- a/tck/features/MatchingSelfRelationships.feature
+++ b/tck/features/MatchingSelfRelationships.feature
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-Feature: SimplePatternSemantics
+Feature: MatchingSelfRelationships
 
   Scenario: Undirected match in self-relationship graph
     Given an empty graph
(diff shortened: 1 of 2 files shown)


boggle:
I think this does not reflect our last discussion on this, does it? Perhaps something worth picking up w Nadime, esp. `MATCH p=(a)-(a)` should match each single self-loop exactly once.

Mats-SX:
@boggle You're right, it doesn't. I'll update it.

Mats-SX:
Updated with latest thinking. Please also see #192.

thobe:
Most of these scenarios are fine. But the ones where you could argue that direction would matter are missing. These are scenarios matching *paths* over loops.

Scenarios like this:

```
Given an empty graph
And having executed:
  """
  CREATE (a)-[:X]->(a)
  """
When executing query:
  """
  MATCH p=()-->()
  MATCH q=()<--()
  RETURN p = q
  """
Then the result should be:
  | p = q |
  |  ???  |
```

thobe:
While I agree with the pattern matching semantics for loops outlined here, I'm not convinced about how this extends to paths in #192, perhaps pattern matching should be considered a separate thing from path matching (at which point perhaps [regular path patterns](https://github.com/opencypher/openCypher/pull/187) should be required for path matching).

PR #166, opened 2016-12-15
Specifying Cypher constraints

Mats-SX:
Specifies syntax for constraints and specifies three concrete constraints: node property uniqueness, node property existence, and relationship property existence.


cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+.Grammar definition for node property uniqueness constraint:
+[source, ebnf]
+----
+uniqueness-constraint = "UNIQUE", property-expression, { ",", property-expression } ;
+----
+
+.Example of single-property uniqueness constraint:
+[source, cypher]
+----
+CREATE CONSTRAINT unique_person_email
```
petraselmer: Are we notionally mandating this particular style (for a constraint name)? Ie. lower case & underscores?

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+FOR (p:Person)
+REQUIRE UNIQUE p.name, p.email, p.address
+----
+
+==== Semantics
+
+The domain of a property uniqueness constraint is defined as all the nodes with a specific label where the specified property key(s) exist (are non-null).
+When more than one property key is defined as part of the constraint, only nodes where _all_ of the properties exist are part of the domain.
+
+The uniqueness mandates that two distinct nodes within the domain can not have the same combination of values for the defined properties (respectively).
```
petraselmer: "uniqueness" -> "uniqueness constraint"
"can not" -> "cannot"

petraselmer: What is meant by 'respectively' in this context?

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+Consider the graph created by the following statement:
+
+[source, cypher]
+----
+CREATE (:Color {name: 'white', rgb: 0xffffff})
+CREATE (:Color {name: 'black', rgb: 0x000000})
+CREATE (:Color {name: 'very, very dark grey', rgb: 0x000000}) // rounding error!
+----
+
+Due to the duplication of the `rgb` property, the following attempt at creating a constraint will fail:
```
petraselmer: "Due" -> "Owing"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+CREATE CONSTRAINT unique_color_nodes
+FOR (c:Color)
+REQUIRE UNIQUE c.rgb, c.name
+----
+
+[[existence]]
+=== Property existence constraints
+
+Property existence constraints are defined for both nodes and relationships, but the semantics are the same.
+For this reason we will go over both constraints in the same section.
```
petraselmer: "For this reason we will go over both constraints in the same section." -> "We now describe both of these."

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+==== Syntax
+
+.Grammar definition for property existence constraint:
```
petraselmer: Add "the"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+[source, ebnf]
+----
+existence-constraint = "exists", "(", property-expression, ")" ;
+----
+
+.Example of node property existence constraint:
```
petraselmer: Add "a"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+[source, cypher]
+----
+CREATE CONSTRAINT colors_must_have_rgb
+FOR (c:Color)
+REQUIRE exists(c.rgb)
+----
+
+.Example of relationship property existence constraint:
```
petraselmer: Add "a"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+[source, cypher]
+----
+CREATE CONSTRAINT rates_have_quality
+FOR ()-[l:RATED]-()
```
petraselmer: Are the `()-` parts required? What about 
`CREATE ... FOR [l:RATED]`?
As it is, I keep thinking the directionality and the node labels are relevant, when they're not

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+REQUIRE exists(l.rating)
+----
+
+==== Semantics
+
+The domain of a node property existence constraint are all nodes with the specified label.
+Similarly, the domain of a relationship property existence constraint are all relationship with the specified type.
+
+Property existence constraints mandates that the value of the specified property exists (is non-null) for all entities in the domain.
```
petraselmer: "Property existence constraints" -> "A property existence constraint"
"(is non-null)" -> "(i.e. is non-null)"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+===== Example
+
+Consider the graph containing `:Color` nodes.
+Each color has an integral RGB value representation in a property `rgb`.
```
petraselmer: Please can we not use "integral"? This is a very overloaded term: intrinsic to something & calculus, which may muddy the waters. We could use it if we define this in a terminology section, but until then, it may not be as clear as it could be. Rather say "of type integer" or similar.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+Users may lookup color nodes to extract their RGB values for application processing.
```
petraselmer: "lookup" -> "look up"

petraselmer: Tighten up "color nodes" -> say nodes labeled with :Color or something like that

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+Users may also add new color nodes to the graph.
```
petraselmer: Same comment as above

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+Suppose the query that looks up the RGB value of a color with a given name looks like this:
```
petraselmer: Same as above

petraselmer: Format "name" so it stands out as an actual thing

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+[source, cypher]
+----
+MATCH (c:Color {name: $name})
+WHERE exists(c.rgb)
+RETURN c.rgb
+----
+
+The `WHERE` clause protects the application from receiving `null` values back for user-defined colors where the RGB values have not been specified correctly.
```
petraselmer: "from receiving `null` values back for" 
->
"from retrieving `null` values for"

petraselmer: "protects" -> "prevents"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+CREATE CONSTRAINT colors_must_have_rgb
+FOR (c:Color)
+REQUIRE exists(c.rgb)
+----
+
+Any updating statement that would create a `:Color` node without specifying a `rgb` property for it would now fail.
+
+=== Interaction with existing features
+
+The main interaction between the constraints and the rest of the language happens during updating statements.
```
petraselmer: "happens" -> "occurs"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+Existing constraints will cause certain updating statements to fail; in fact, that's the main purpose.
```
petraselmer: "; in fact, that's the main purpose"
->
" - which is, in fact, the main purpose of this feature"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+=== Alternatives
+
+Plenty of alternative syntaxes have been discussed:
```
petraselmer: "Plenty of" -> "Many"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+* `GIVEN`, `CONSTRAIN`, `ASSERT` instead of `FOR`
+* `ASSERT`, `ENFORCE`, `IMPLIES` instead of `REQUIRE`
+
+The use of existing expression to express uniqueness, instead of using a new keyword `UNIQUE`, on the form:
```
petraselmer: "of existing expression" -> add an article

petraselmer: "on the form" -> do you mean "of the form"?

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+----
+FOR (p:Person), (q:Person)
+REQUIRE p.email <> q.email AND p <> q
+----
+which quickly becomes unwieldy for multiple properties.
```
petraselmer: I think it's more readable if this sentence is placed _before_ the code snippet

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+== What others do
+
+In SQL, the following constraints exist (http://www.w3schools.com/sql/sql_constraints.asp):
+
+* `NOT NULL` - Indicates that a column cannot store NULL value
```
petraselmer: "a NULL value"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+* `UNIQUE` - Ensures that each row for a column must have a unique value
+* `PRIMARY KEY` - A combination of a `NOT NULL` and `UNIQUE`. Ensures that a column (or combination of two or more columns) have a unique identity which helps to find a particular record in a table more easily and quickly
```
petraselmer: "have a unique" -> "has a unique"

petraselmer: "which helps to find a particular record in a table more easily and quickly"
->
", reducing the resources required to locate a specific record in a table"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+* `FOREIGN KEY` - Ensure the referential integrity of the data in one table to match values in another table
```
petraselmer: "Ensure the referential integrity of the data in one table to match values in another table"
->
"Ensures the referential integrity of the data in one table matches values in another table"

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+* `CHECK` - Ensures that the value in a column meets a specific condition
+* `DEFAULT` - Specifies a default value for a column
+The next chapters will describe each constraint in detail.
+
+The property existence constraints represent the same functionality as the `NOT NULL` SQL constraint.
+The node property uniqueness constraint represents the `PRIMARY KEY` SQL constraint.
+
+SQL constraints may be introduced at table creation time (in a `CREATE TABLE` statement), or in an `ALTER TABLE` statement:
+
+.Creating a persons table in SQL Server / Oracle / MS Access:
```
petraselmer: Format 'persons'

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+[source, sql]
+----
+CREATE TABLE Persons
+(
+   P_Id int NOT NULL UNIQUE,
+   LastName varchar(255) NOT NULL,
+   FirstName varchar(255))
+----
+
+.Creating a persons table in MySQL:
```
petraselmer: Format 'persons'

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+.Adding a named composite `UNIQUE` constraint in MySQL / SQL Server / Oracle / MS Access:
+[source, sql]
+----
+ALTER TABLE Persons
+ADD CONSTRAINT uc_PersonID UNIQUE (P_Id,LastName)
+----
+
+== Benefits to this proposal
+
+Constraints make Cypher's notion of schema more well-defined, and allows users to keep graphs in a more regular, easier to manage form.
```
petraselmer: "keep graphs in a more regular, easier to manage form."
->
"maintain graphs in a more regular, easier-to-manage form."

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,285 @@
...
+
+== Caveats to this proposal
+
+For an implementing system, some constraints may prove challenging to enforce, as they generally require scanning through large parts of the graph to look for conflicting entities.
```
petraselmer: "For an implementing system"
->
"For a system seeking to implement the contents of this CIP"

petraselmer: "look for" -> "locate"

thobe:
I thought the idea was that we were going to specify a syntax for constraints without specifying which particular constraints an implementation should support. The syntax definition here explicitly talks only about `uniqueness-constraint` and `existence-constraint`.

Mats-SX:
CIP has now been reworked to try and fit the model discussed.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,292 @@
...
+
+The constraint syntax is defined as follows:
+
+.Grammar definition for constraint syntax.
+[source, ebnf]
+----
+constraint command = create-constraint | drop-constraint ;
+create-constraint  = "CREATE", "CONSTRAINT", constraint-name, "FOR", constraint-pattern, "REQUIRE", constraint-expr, { "REQUIRE", constraint-expr } ;
+constraint-name    = symbolic-name
+constraint-pattern = node-pattern | simple-pattern ;
```
boggle: Or even a comma-separated sequence?

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,310 @@
...
+Any updating statement that would create a `:Color` node without specifying an `rgb` property for it would now fail.
+
+Alternatively, we could extend our previous constraint definition with this new requirement:
+
+[source, cypher]
+----
+CREATE CONSTRAINT color_schema
+FOR (c:Color)
+REQUIRE UNIQUE c.rgb, c.name
+REQUIRE exists(c.rgb)
```
boggle: I'm missing an example for an existence constraint that requires multiple properties to exist on a node; this can be expressed using `exists(n.foo) AND exists(n.bar)`.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,310 @@
...
+
+== What others do
+
+In SQL, the following constraints exist (inspired by http://www.w3schools.com/sql/sql_constraints.asp):
+
+* `NOT NULL` - Indicates that a column cannot store a null value.
+* `UNIQUE` - Ensures that each row for a column must have a unique value.
+* `PRIMARY KEY` - A combination of a `NOT NULL` and `UNIQUE`. Ensures that a column (or a combination of two or more columns) has a unique identity, reducing the resources required to locate a specific record in a table.
+* `FOREIGN KEY` - Ensures the referential integrity of the data in one table matches values in another table.
+* `CHECK` - Ensures that the value in a column meets a specific condition
```
boggle: Maybe we should have a leading keyword like that for non-UNIQUE constraints as well?

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
Mats-SX: You're right, I don't see why not.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
Mats-SX: Right, or using multiple `REQUIRE`. I can add both.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,292 @@
...
+This allows for very complex concrete constraint definitions within the specified syntax.
+
+To that set of valid expressions, this CIP further specifies a special prefix operator `UNIQUE`, which is used to assert uniqueness of one or more property expressions (see <<uniqueness>> for details).
+
+==== Constraint names
+
+All constraints require the user to specify a nonempty _name_ at constraint creation time.
+This name is subsequently the handle with which a user may refer to the constraint, for example when dropping it.
+
+// TODO: Should we impose restrictions on the domain of constraint names, or are all Unicode characters allowed?
```
Mats-SX: Escaping is captured by referring to the `symbolicString` rule.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,350 @@
...
+The constraint syntax is defined as follows:
+
+.Grammar definition for constraint syntax.
+[source, ebnf]
+----
+constraint command    = create-constraint | drop-constraint ;
+create-constraint     = "CREATE", "CONSTRAINT", [ constraint-name ], "FOR", pattern, "REQUIRE", constraint-predicate, { "REQUIRE", constraint-predicate } ;
+constraint-name       = symbolic-name
+constraint-predicate  = expression | unique | primary-key ;
+uniqune               = "UNIQUE", property-expression
```
IanRogers: typo? uniqune -> unique

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,350 @@
...
+REQUIRE p:Person
+----
+
+.Constraint example from CIR-2017-173
+[source, cypher]
+----
+CREATE CONSTRAINT enforce_dag_unique_for_R_links
+FOR p = (a)-[:R]-(b)
+REQUIRE length(p) < 2
+----
```
IanRogers: Sorry, this isn't what https://github.com/opencypher/openCypher/issues/173 is talking about. The "unique" constraint on a DAG would be `require size(p) < 2` though this is a generalisation of `require unique p` (i.e. to enforce the behaviour of `create unique` http://neo4j.com/docs/developer-manual/current/cypher/clauses/create-unique/#_create_unique_relationships)

Mats-SX: You're right; `length()` actually doesn't make much sense here (it will always be 1 with this pattern). Although `size()` doesn't either, as it's not defined over paths at all.

Perhaps 

```
REQUIRE size(collect(DISTINCT p)) < 2
```

would work, but it's not very pretty. We probably need some other construct here. I'll remove this example for now.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
Mats-SX: Thanks!

thobe:
`PRIMARY KEY` is not a helpful name for the concept it is used for describing.

On the high level there are two reasons for this:

1. The word _"primary"_ does not mean anything that is helpful in this context.
2. The concept of _primary keys_ carries with it a lot of associations from relational databases, many of which do not apply to the property graph model.
   * it should be noted that for the relational database case the word "primary" _does_ have relevant meaning.

Diving further into these reasons, starting with the word _"primary"_:

* It implies that there is such a thing as a _"secondary"_ key as well. In a relational database a secondary key is any index on a table other than the primary key.
* It implies that this key has higher importance than any other key. While this might be true in many actual domain models, it is not always the case - in some cases there are other keys of equal importance.
* It carries with it the association that there can be only one _primary key_. In many implementations - Neo4j being one of them - there is absolutely no need to enforce such a constraint on the ability to model your data.
* It implies that the key needs to be defined first - before any data is inserted. In many implementation - Neo4j being one of them - this is not the case.

As for the aspects of _primary keys_ in relational databases that do not apply to the property graph model:

* The notion of there being a _primary key_ implies that there might also be a _foreign key_ - the idea of having _foreign keys_ in a graph is quite silly, since we have direct relationships.
* Coming from relational databases, I would expect preferential treatment for primary keys over any other (secondary) keys. I would expect lookup based on the primary key to be faster than any other key, since I would expect the data in that table to be structured by the primary key. In essence I would expect that leaf nodes in the index for the primary key being the actual row, with all of its data. Whereas for a secondary key the leaf node would only point to the actual row in the primary structure - there would be indirection for accessing the full data by a secondary key, thus penalizing access by secondary key.
  * Again this would not be true in for example Neo4j, where every key is actually secondary.
* Relational databases equate the _primary key_ with _identity_, the property graph (at least in some implementations - for example Neo4j) has a separate notion of identity, and while the type of key we are proposing to add to the model would allow you to uniquely identify an entity, it would not necessarily identify that same entity forever - the same entity might change some of the values of the key and thus be identified by a different key, but still have the same identity.

The only reason I can think of for introducing the concept of a _primary key_ in Cypher is for being able to map Cypher onto a relational database model. If that is the case I would much rather see this proposed from a vendor working on such a mapping, since they would have the insight into what needs to be modeled.

I do think that the notion of a unique indexed key of mandatory properties is helpful, and I see the benefit of elevating such a concept to the status of receiving its own syntax, but I don't think `PRIMARY KEY` is a good name for it.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,342 @@
...
+
+.Example of a constraint definition equivalent to the above `PRIMARY KEY` example:
+[source, cypher]
+----
+CREATE CONSTRAINT person_details
+FOR (p:Person)
+REQUIRE UNIQUE p.name
+REQUIRE UNIQUE p.email
+REQUIRE UNIQUE p.address
+REQUIRE exists(p.name) AND exists(p.email) AND exists(p.address)
```
thobe: This decomposition seems wrong to me, we would want the compound key to require that the _combination_ of `p.name`, `p.email`, and `p.address` is unique, not that each of them is unique on their own. But we do require that they each exist.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,342 @@
...
+
+=== Syntax
+
+The constraint syntax is defined as follows:
+
+.Grammar definition for constraint syntax.
+[source, ebnf]
+----
+constraint command    = create-constraint | drop-constraint ;
+create-constraint     = "CREATE", "CONSTRAINT", [ constraint-name ], "FOR", pattern, "REQUIRE", constraint-predicate, { "REQUIRE", constraint-predicate } ;
```
thobe: I know there has been conversation about how `CREATE` is an overloaded word if used for constraints as well, since it is already used for creating nodes and edges. I just wanted to plant the idea of using the word `DEFINE` for constraints instead. i.e. `DEFINE CONSTRAINT FOR (p:Person) REQUIRE UNIQUE p.firstName, p.lastName`

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,342 @@
...
+constraint-name       = symbolic-name
+constraint-predicate  = expression | unique | primary-key ;
+unique                = "UNIQUE", property-expression
```
thobe: I think we want this to be a list of expressions in order to allow compound keys to be unique.

Mats-SX: I view that as out of scope for the first version of this CIP.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,342 @@
...
+
+* Attempting to create a constraint on a graph where the data does not comply with the constraint criterion.
+* Attempting to create a constraint with a name that already exists.
+* Attempting to drop a constraint referencing a non-existent name.
+* Attempting to modify the graph in such a way that it would violate a constraint.
+
+==== Mutability
+
+Once a constraint has been created, it may not be amended.
+Should a user wish to change its definition, it has to be dropped and recreated with an updated structure.
```
thobe: Should we have a note here that transactional implementations could do both the dropping and recreation in the same transaction so that the constraint is atomically mutated? This would of course allow leaving the old constraint in place should the creation of the new constraint fail.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
Mats-SX: Right, without `UNIQUE` on multiple properties, this is not expressible in another way.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
```diff
@@ -0,0 +1,342 @@
...
+3. [[domain-exception]]Entities for which a constraint expression evaluate to `null` under Cypher's ternary logic are _excluded_ from the constraint domain, even if they fit within the constraint pattern.
+
+==== Errors
+
+The following list describes the situations in which an error will be raised:
+
+* Attempting to create a constraint on a graph where the data does not comply with the constraint criterion.
+* Attempting to create a constraint with a name that already exists.
+* Attempting to drop a constraint referencing a non-existent name.
+* Attempting to modify the graph in such a way that it would violate a constraint.
```
thobe: `+` attempting to create a constraint that the underlying engine does not support enforcing.

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
thobe: if so, I would propose the word `THAT`, since it reads nicely...

```
CREATE CONSTRAINT FOR (x:Foo) REQUIRE THAT ...
```

thobe:
I wonder if the notion of _unique key_ (that at the moment are erroneously called _primary key_) should really be a constraint, or if it should have its own syntax, something like:

```
CREATE KEY FOR (p:Person) AS p.name, p.address
```

The reason for this being that it actually implies multiple constraints, and typically also an index. Since it is a composed concept like that, perhaps it would be sensible to elevate it to being a syntactical concept of its own.

In the syntax for this, if accepted, we should allow an optional name for the key as well, just like we do for constraints.

Mats-SX:
The CIP now uses `ADD`, `NODE KEY` and details a return record. I also took several review comments into account (thanks!).

cip/1.accepted/CIP2016-12-14-Constraint-syntax.adoc
Mats-SX: No, it is just an example. It may be any symbolic string.

Mats-SX:
It is a little difficult to manage the different comments in the review if they are all in the same message, instead of comments made on/near the lines that triggered them. That would help discussions to be kept disentangled, and suggestions such as the one in item `5.` would be easier to adopt directly. In the absence of this, I will try to address comments in a master-comment following the same numbering scheme:

1. I know nothing of that CIP and can not comment on it specifically. Non-alignment is of course undesirable, but I would favour progress over perfection (to steal a quote) and accept this CIP until there's a conflict identified, and deal with it at that point.

2. Fixed.

3. It is not a syntax rule, it is a disclaimer which intends to note that the Examples are purely informative and only pose examples on possible constraints. If you think that is self explanatory by the fact that these are examples we can scratch the note entirely. I'm not sure adding a note in the syntax rules saying that an implementation may omit support for certain patterns or predicates or combinations thereof would improve this specification; surely any implementer may impose any implementation-specific limitations they would see fit anyhow?

4. Added a 4th point under `Semantics`.

5. I've added a section `Plurality`, extended the syntax definition and added an example. However, I am not sure comma would be the appropriate symbol to use as a separator; it is not significant enough to stick out between declarations.

6. I'm not so keen on `CHECK` as I rather like `REQUIRE`. In a way one could say that this is the name of the 'standard' constraints. Could make it optional for `UNIQUE` and `KEY` although I think it inhibits readability to omit it.

7. I added an explanatory paragraph in the Syntax section to clarify.

----

At this point we discussed this CIP in the CLG, mostly focusing on points `1.` and `6.` above, and decided to pause the work on this CIP awaiting maturity on the schema topic.

Mats-SX:
This CIP has now been updated. See commit history for a summary of new models.

TCK tests cannot be added as the CIP mandates no particular concrete constraint predicates to be enforced by every implementation of Cypher. However, I will exemplify TCK scenarios that one could consider if one were to implement the NODE KEY constraint:

```
Feature: CreateConstraint1

  Scenario: [1] Blocking creation of nodes that do not conform to NODE KEY constraint
    Given an empty graph
    And having executed:
      """
      CREATE CONSTRAINT
      FOR (a:A)
      REQUIRE (a.x, a.y) IS NODE KEY
      """
    When executing query:
      """
      CREATE (a:A)
      SET a.x = 1
      """
    Then ConstraintValidationFailed should be raised at runtime: NodeKeyRequired

  Scenario: [2] Allowing creation of nodes that uphold NODE KEY constraint
    Given an empty graph
    And having executed:
      """
      CREATE CONSTRAINT
      FOR (a:A)
      REQUIRE (a.x, a.y) IS NODE KEY
      """
    When executing query:
      """
      CREATE (a:A)
      SET a.x = 1
      SET a.y = 2
      """
    Then the result should be empty
    And the side effects should be:
      | +nodes      | 1 |
      | +properties | 2 |
```

PR #167, opened 2016-12-19
Reserved words

Mats-SX:
Cypher should specify a set of reserved words that may not be used as variables, parameters, or function names. This would make parsing the language simpler, make it harder to write particularly confusing statements, and is in alignment with how computer languages in general are specified.

### DIFF
--- a/cip/1.accepted/CIP2016-12-19-Reserved-keywords.adoc
+++ b/cip/1.accepted/CIP2016-12-19-Reserved-keywords.adoc
@@ -1,0 +1,163 @@
+= CIP2016-12-19-Reserved-keywords
+:numbered:
+:toc:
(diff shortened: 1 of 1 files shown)


FylmTM:
I really like this CIP.
From experience working with Cypher grammar - allowing keywords to be used as variable names is really pain to work with.
If parser is not build with PEG (e.g. BNF), then it is required to add hacks into a grammar to allow keywords to be used as variables. Then, to provide proper support (e.g. syntax highlight) you need to do post-processing to mark such keywords as valid names.

I should add that keywords should not be allowed in:
- Variables
- Labels
- Relationship types
- Function names
- Procedure names
- User function / Stored procedure output parameters
- Parameters

Issue #179, opened 2017-02-03
Add support for Conjunctive Regular Path Queries (CRPQs)


thobe:
In order to support expressing more complex patterns over the graph, a syntax for Conjunctive Regular Path Queries would be desirable. Such a feature should support the following combinations of path patterns:

- Alternatives (or unions) (_either this pattern or that pattern_)
- Sequences (_first this pattern, than that pattern_)
- Transitive closure (_this pattern repeated multiple times_)
- Predicates on node and relationship properties.

Possibly we might also want a way to express _any path *not matching* the given pattern_.

petraselmer:
(Updated the original comment to include the name of the CIR)

thobe:
I have started working on a proposal for how to bring Regular Path Queries to Cypher. See [CIP2017-02-06](https://github.com/opencypher/openCypher/pull/187).

PR #187, opened 2017-02-06
Regular Path Queries

thobe:
This is first draft of a proposal for adding _Regular Path Patterns_ to Cypher. There is still work to be done here before this is finalised.


thobe:
This proposal should eventually be able to fulfil the requirements outlined in #179.

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
```diff
@@ -0,0 +1,160 @@
...
+.Finding someone loved by someone hated by someone you know, transitively
+----
+MATCH (you)-/(:KNOWS :HATES)+ :LOVES/->(someone)
+----
+
+Note the `+` expressing one or more occurrences of the sequence `KNOWS` followed by `HATES`.
+
+The direction of each relationship is governed by the overall direction of the Regular Path Pattern.
+It is however possible to explicitly define the direction for a particular part of the pattern.
+This is done by either prefixing that part with `<` for a right-to-left direction or suffixing it with `>` for a left-to-right direction.
```
Mats-SX: We could use 'outgoing' instead of 'left-to-right', and 'incoming' instead of 'right-to-left'.

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
```diff
@@ -0,0 +1,160 @@
...
+.Find friends of friends that are not haters
+----
+MATCH (you)-/:KNOWS not_a_hater :KNOWS/-(friendly_friend_of_friend)
+PATH (x)-/not_a_hater/-(x) IS (x)
+     WHERE NOT EXISTS { (x)-[:HATES]->() }
+----
+
+In the case of a Defined Path Predicate where both nodes are the same, the direction of the predicate is irrelevant.
+In general the direction of a Defined Path Predicate is quite important, and used for mapping the pattern in the predicate into the Regular Path Patterns that reference it.
+The only cases where it is allowed to omit the direction of a Defined Path Predicate is when the defined predicate is reflexive.
```
Mats-SX: Will it be the job of the query compiler to determine reflexiveness of the defined predicate, and issue an error if non-reflexive predicates are defined (or used) without direction?

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
```diff
@@ -0,0 +1,160 @@
...
+[source, cypher]
+.Find friends of friends that are not haters
+----
+MATCH (you)-/:KNOWS not_a_hater :KNOWS/-(friendly_friend_of_friend)
+PATH (x)-/not_a_hater/-(x) IS (x)
+     WHERE NOT EXISTS { (x)-[:HATES]->() }
+----
+
+In the case of a Defined Path Predicate where both nodes are the same, the direction of the predicate is irrelevant.
+In general the direction of a Defined Path Predicate is quite important, and used for mapping the pattern in the predicate into the Regular Path Patterns that reference it.
```
Mats-SX: This applies to the definition of a predicate, right? I imagine that it will be perfectly possible to reverse the direction in the Regular Path Pattern, like so:

```
MATCH (a)<-/pred/-(b)
PATH (a)-/pred/->(b) IS
     (a)-[:KNOWS]->(b)
```

How about it being used in an undirected Regular Path Pattern?

```
MATCH (a)-/pred/-(b)
PATH (a)-/pred/->(b) IS
     (a)-[:KNOWS]->(b)
```

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
thobe: That would imply a particular traversal order, which I think is saying too much, with `left-to-right` and `right-to-left` we are only talking about the direction with regard to how the pattern is _written_.

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
thobe: Both of those are perfectly valid. The undirected use of a directed defined path predicate is the same as an OR/UNION between the two directions.

I.e. `(a)-/pred/-(b)` is the same as `(a)-/<pred | pred>/-(b)` (which is the same as `(a)-/pred> | <pred/-(b)` or `(a)-/<pred>/-(b)` or `(a)<-/pred/->(b)`).

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
thobe: I'd say that either the query compiler issues an error, _or_ the behaviour of the path predicate is undefined (i.e. the behaviour in such a case is outside of the scope of the specification).

I'd prefer for an implementation to issue an error, but I don't know how hard that analysis is to perform, and I thus don't want to mandate that at this point (it needs further analysis).

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
```diff
@@ -0,0 +1,160 @@
...
+:source-highlighter: codemirror
+
+*Authors:* Tobias Lindaaker <tobias.lindaaker@neotechnology.com>
+
+toc::[]
+
+== Regular Path Patterns
+
+Above and beyond the types of patterns that can be expressed in Cypher using the normal path syntax, Cypher also supports what amounts to regular expressions over paths.
+This functionality is called Regular Path Patterns.
```
petraselmer: I think it would be a good idea to mention somewhere that these are essentially RPQs, since RPQs are the standard term for these sorts of queries.

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
```diff
@@ -0,0 +1,160 @@
...
+
+A Regular Path Pattern is defined as:
```
petraselmer: For 1 - 4 below, it would be great if these could be supplemented with regex notation

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
```diff
@@ -0,0 +1,160 @@
...
+• A Regular Path Pattern followed by another Regular Path Pattern, or
+• An alternative between two Regular Path Patterns, or
+• A repetition of a Regular Path Pattern, or
+• A reference to a Defined Path Predicate.
+
+Regular Path Patterns are written similarly to how relationship patterns are written, but enclosed within two slash (`/`) characters instead of brackets (`[]`).
+
+Contrary to Relationship Patterns, Regular Path Patterns do _not_ allow binding a relationship to a variable.
+In order to bind the matching path to a variable, a Path Assignment should be used, by preceding the path with an identifier and an equals sign (`=`).
+This avoids a problem that existed in the past with repetition of relationships (a syntax that was deprecated with the introduction of Regular Path Patterns), where a relationship variable would bind to a list, making it hard to express predicates over the actual relationships.
```
petraselmer: Is it not meant to be "a syntax that _is_ deprecated" (as RPPs are only being introduced now)?

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
thobe: As in an example of what the syntax looks like?

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
thobe: I've tried writing the text as if Regular Path Patterns are already in the language, since when this text is merged, it will be in the language. Although _is_ works in that case as well.

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
```diff
@@ -0,0 +1,160 @@
...
+----
+MATCH (you)-/(:KNOWS :HATES)+ :LOVES/->(someone)
+----
+
+Note the `+` expressing one or more occurrences of the sequence `KNOWS` followed by `HATES`.
+
+The direction of each relationship is governed by the overall direction of the Regular Path Pattern.
+It is however possible to explicitly define the direction for a particular part of the pattern.
+This is done by either prefixing that part with `<` for a right-to-left direction or suffixing it with `>` for a left-to-right direction.
+It is possible to both prefix the part with `<` and suffixing it with `>`, giving that part the interpretation of being undirected.
```
petraselmer: "suffixing" -> "suffix"

petraselmer: The "giving that part the interpretation of being undirected" section is a bit unwieldy - maybe something along the lines of "indicating that that part of the pattern is undirected"

cip/1.accepted/CIP2017-02-06-Regular-Path-Patterns.adoc
petraselmer: Aha - ok, that makes sense

thobe:
I've updated the document to unify the Path Pattern syntax with normal Pattern syntax. I have not yet updated the grammar to reflect that, I'll do that imminently.

thobe:
I think it is important to note which things that is currently valid Cypher will have changed semantics under this proposal:

* Binding a variable under repetition will no longer be possible. I.e. this pattern is now invalid: `(a)-[rels*]->(b)`
* Property predicates will have a different scope if colons are used when separating alternatives of relationship types. The pattern `()-[:FOO|:BAR{baz=17}]-()` used to mean _any relationship of type `FOO` or type `BAR` where the property `baz` has value `17`_, but under this proposal the semantics would be _any relationship of type `FOO` or any relationship of type `BAR` where the property `baz` has value `17`_, i.e. under the old semantics the property predicate applied to both relationship types, but under the new semantics it only applies to the `BAR` relationships. The old semantics are still used if colon is not used in the listing of relationship types as in `()-[:FOO|BAR{baz=17}]-()`.

thobe:
Notable example queries that *cannot* be expressed using this syntax includes:

- [ ] Paths between two nodes that _does not_ match a given pattern.
    Cypher can still express such a _cartesian product_, but not as a path.
- [ ] Paths where the value of a given property in all nodes of the path differs from the value of that property at the first node of the path.
    This is one of the canonical examples of _Regular Expressions with Memory_ (REMs)
- [ ] Paths where a certain property has a different value in all nodes.

cip/1.accepted/CIP2017-02-06-Path-Patterns.adoc
```diff
@@ -0,0 +1,439 @@
...
+
+In the case of a Defined Path Predicate where both nodes are the same, the direction of the predicate is irrelevant.
+In general the direction of a Defined Path Predicate is quite important, and used for mapping the pattern in the predicate into the Path Patterns that reference it.
+The only cases where it is allowed to omit the direction of a Defined Path Predicate is when the defined predicate is reflexive.
+This is obviously the case when both nodes are the same, but it would also be the case when the internal pattern is symmetrical, such as in the following example:
+
+[source, cypher]
+.Find chains of co-authorship
+----
+PATH PATTERN co_author = (a)-[:AUTHORED]->(:Book)<-[:AUTHORED]-(b)
```
thobe: But it is an example of when the named path pattern _itself_ is undirected - and the direction *is* left out on the next row.

Mats-SX: Ah, so the meaning is that it is allowed to omit the direction of a _path pattern_ that _references_ a DPP, but only when the DPP is reflexive. Got it.

petraselmer:
Two things:

1) Please can some more examples be added to further exemplify the use of this feature?
 
For example, drawn from life sciences, it would be good to see these queries:
Query 1) 'Return all pairs of directly-connected nodes (a,b) where every second node in the path must have label A.' 
Query 2) 'Return all pairs of directly-connected nodes (a,b) where there are at least 2 instances of a node labelled with X linked to a node labelled with Y in the path.'

2) Your comment beginning '_Notable example queries that cannot be expressed using this syntax includes...._' is very pertinent. Can this be added to the CIP itself for the purposes of rigour and completeness?

cip/1.accepted/CIP2017-02-06-Path-Patterns.adoc
```diff
@@ -246,6 +246,61 @@ PATH PATTERN co_author = (a)-[:AUTHORED]->(:Book)<-[:AUTHORED]-(b)
...
+MATCH (a), (b)
+WHERE NOT EXISTS { (a)-/𝜶/-(b) }
+----
+
+Queries like this are generally not tractable, so arguably it is a good thing that they are not easy to express.
+
+==== Differing property values along a path
+
+While it is possible to express that a certain property should have the same value for all nodes in a path (by saying that each pair of nodes should have the same property value), it is not possible to express that all nodes should have a _different_ property value.
+It has been shown that computing such paths would not be tractable in the general case, so perhaps it is a good thing to not be able to express this.
```
Mats-SX: Perhaps we should add a reference to this claim.

thobe:
I wrote up some slides that give an overview of the history of this proposal, what is has been influenced by and what other things has been influenced by it. This information should provide some insight into some of the design choices made in this CIP.

The slides are available on the [opencypher.org/references page](https://s3.amazonaws.com/artifacts.opencypher.org/website/materials/Comparing+Path+Pattern+Syntaxes.pdf)

cip/1.accepted/CIP2017-02-06-Path-Patterns.adoc
```diff
@@ -0,0 +1,503 @@
...
+
+RangeDetail = [IntegerLiteral | Parameter], '..', [IntegerLiteral | Parameter] ;
+----
+
+The `PathReference` is a reference to a Named Path Predicate.
+These are defined using the following syntax:
+
+[source, ebnf]
+----
+NamedPathPredicate   = 'PATH', 'PATTERN', NamedPathName, '=', PathPattern, [Where] ;
```
gsvgit, line 105: It looks like examples are not match with grammar for ```NamedPathPredicate```. In examples like 
```
PATH PATTERN unreciprocated_love = (a)-[:LOVES]->(b)
```
we see, that it should be something like ```PatternPart``` (maybe without ```[Variable, '=']``` part) in ```NamedPathPredicate``` after ```=``` instead of ```PathPattern```.

So, I think that correct rule should be 
```
NamedPathPredicate   = 'PATH', 'PATTERN', NamedPathName, '=', NodePattern, {(EdgePattern | PathPattern), NodePattern}, [Where] ;
```

thobe: Yes, you are right, thank you!

PR #192, opened 2017-02-27
Change definition of path equality

Mats-SX:
Undirected matching on self-relationships (aka loops) is tricky. #162 makes an effort to try and make explicit the intended semantics. This leaves paths in an ill-defined state, however. Consider the graph with one node and one relationship (to itself), and the following query:

```
MATCH p = ()--()
RETURN p
```

#162 states that only one row is to be returned (there's only one combination of node-rel-node available in the data graph). However, if paths are supposed to capture the direction in which they traverse relationships, then which path should be returned by this query: the one that traverses the relationship outgoing, or the one that does it incoming? It seems ill-defined.

In no other circumstance other than a self-relationship will the traversed direction matter.

Same idea as in #160

### DIFF
--- a/tck/features/PathEquality.feature
+++ b/tck/features/PathEquality.feature
@@ -1,0 +1,35 @@
+#
+# Copyright 2017 "Neo Technology",
+# Network Engine for Objects in Lund AB (http://neotechnology.com)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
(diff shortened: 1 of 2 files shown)


tck/features/PathEquality.feature
```diff
@@ -0,0 +1,35 @@
...
+      """
+    When executing query:
+      """
+      MATCH p1 = (:A)-->()
+      MATCH p2 = (:A)<--()
+      RETURN p1 = p2
+      """
+    Then the result should be:
+      | p1 = p2 |
+      | true    |
```
thobe: This feels wrong to me.

thobe:
I think there are more scenarios that should be considered, for completeness:

```
Given an empty graph
And having executed:
  """
  CREATE (a)-[:X]->(a)
  """
When executing query:
  """
  MATCH p=()-->()
  MATCH q=()--()
  RETURN p = q
  """
Then the result should be:
  | p = q |
  |  ???  |
```

Where the result could be either:
```
  | p = q |
  | true  |
```
or:
```
  | p = q |
  | true  |
  | false |
```
(without order)

and

```
Given an empty graph
And having executed:
  """
  CREATE (a)-[:X]->(a)
  """
When executing query:
  """
  MATCH p=()<--()
  MATCH q=()--()
  RETURN p = q
  """
Then the result should be:
  | p = q |
  |  ???  |
```

With the same possible results as above.

PR #197, opened 2017-03-03
Add Neo4j index extension CIP

Mats-SX:
This reflects a Neo4j extension of the constraints syntax suggested in #166 to support index creation/deletion.

[Direct link to CIP](https://github.com/Mats-SX/openCypher/blob/neo4j-indexes/cip/vendor-extensions/neo4j/CIP2016-12-14-Neo4j-indexes.adoc)

### DIFF
--- a/cip/vendor-extensions/neo4j/CIP2016-12-14-Neo4j-indexes.adoc
+++ b/cip/vendor-extensions/neo4j/CIP2016-12-14-Neo4j-indexes.adoc
@@ -1,4 +1,4 @@
-= CIP2016-12-16 - Neo4j Indexes
+= CIP2016-12-16 Neo4j Indexes
 :numbered:
 :toc:
 :toc-placement: macro
--- a/cip/vendor-extensions/neo4j/CIP2016-12-14-Neo4j-indexes.adoc
+++ b/cip/vendor-extensions/neo4j/CIP2016-12-14-Neo4j-indexes.adoc
@@ -26,7 +26,7 @@ These nodes are kept in a separate, persisted data structure which allows lookup
 
 === Syntax
 
-The index syntax is based on the constraint syntax, and is detailed below:
+The index syntax is based on the constraint syntax (see the Constraint Syntax CIP), and is detailed below:
 
 .Grammar definition for Neo4j index syntax.
 [source, ebnf]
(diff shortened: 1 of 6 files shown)


cip/vendor-extensions/neo4j/CIP2016-12-14-Neo4j-indexes.adoc
```diff
@@ -0,0 +1,138 @@
...
+The `index-key` expression defines the key for the index, and consist of one or more property expressions, which refer to the entity defined in the pattern.
+
+==== Index names
+
+Just like constraints, indexes have names.
+Neo4j does not support user-defined names, but the index will be assigned a system-generated name.
+
+==== Removing indexes
+
+An index is removed by referring to its name.
```
systay: Since the name is optional, how does one drop an index that is not named?

Mats-SX: The system will generate a name. Only the 'user-defined' part is optional; they should always have a name.

Mats-SX:
This CIP is now also ready for review. It builds on top of #166 so there's no need to review changes that come from that PR here. The reason to build it on top is because this CIP makes many references to the general Constraint Syntax CIP.

sherfert:
I have some questions:

* Having `node-pattern` in the grammar, would that not allow writing `CREATE INDEX foo FOR (a:Foo:Bar) ON a.prop`? I.e. I think we cannot allow multiple labels or no labels, or if we do, we should explain that neo4j does not currently support that and what the semantics are.

* I know that neo4j does not store null-properties in indexes, but is there a fundamental reason why we would want to enforce that? I know we have been discussing null-indexes for index-backed-order-by, so I think it can be a valuable thing that we do not want to forbid.

* Is there a recommendation on how systems should generate unique names for constraints/indexes? I think that would be valuable.

Mats-SX:
CLG review finds on @sherfert's points:

- We'll go with general syntax for patterns and implementation limitations
- Explicitly forbidding `null` values will be removed from this CIP
- No recommendation is required

Mats-SX:
From LangStar review:

- Extend with `OPTIONS` syntax to specify particular index details (such as implementation type)
- Allow not only node patterns, but a general pattern including relationships
- Require property expressions to be delimited using parentheses

PR #204, opened 2017-03-07
Use explicit IDs for longer-lived link targets

jjaderberg:
Generated IDs for section headers and other referenceable content are fragile as they may change unexpectedly.
Explicitly assigned ID are more stable.


petraselmer:
Thanks @jjaderberg - this is great

Issue #177, opened 2017-02-03
Make it easier collect the result of a matched pattern


thobe:
Something like a shorthand syntax for a subquery that combines (`OPTIONAL`) `MATCH`, `collect(...)` and `RETURN` would be really useful, in particular for creating object hierarchy projections from the graph.

One thought would be to use something like a comprehension syntax for this. Something like this:

```
MATCH (user:Person{username:$username})
RETURN [
    MATCH (user)-[:FRIEND]-(friend)
    RETURN friend.username
] AS friends
```

Would be a shorthand for:

```
MATCH (user:Person{username:$username})
OPTIONAL MATCH (user)-[:FRIEND]-(friend)
WITH user, collect(friend.username) AS friends
RETURN friends
```

This particular example is deliberately kept simple, in order to capture only the essence of the proposal, but even if the syntactic saving isn't very substantial, there is a clear benefit to having the scope delineation take care of the grouping of friends per user that the superfluous variable in the aggregating `WITH` statement is otherwise needed for.

Mats-SX:
_Pattern comprehension_ syntax as implemented in Neo4j:

```
MATCH (user:Person {username: $username})
RETURN [(user)-[:FRIEND]-(friend) WHERE friend.age > $age | friend.username] AS friends
```

petraselmer:
(Updated the original comment with the name of the CIR)

PR #217, opened 2017-04-13
CIP2017-03-29 Scalar Subqueries and List Subqueries

thobe:
Solves #177.


thobe:
There is still some work to be done here, but comments are welcome.

cip/1.accepted/CIP2017-03-29-Single-Value-Subqueries.adoc
```diff
@@ -0,0 +1,68 @@
...
+List Subqueries are read-only subqueries that produce a single value per row, and zero or more rows.
+The result of a List Subquery is to collect the value of all rows produced by the subquery into a list.
+
+=== Syntax
+
+[source, ebnf]
+----
+ListSubquery = '[', SingleValueSubquery, ']' ;
+
+ScalarSubquery = '<(((<', SingleValueSubquery, '>)))>' ;
```
thobe: These fish-brackets are a place holder, we really need something better to delineate a scalar subquery. Suggestions are most welcome. It could be a keyword, I considered `'SINGLE', '(', SingleValueSubquery, ')' ;`, but `single` is already used as a predicate function (actually also a type of subquery) in Cypher.

Another suggestion was `'SCALAR', '(', SingleValueSubquery, ')' ;`, but that didn't feel quite right.

boggle: More random ideas:

- `VALUE ( subquery )` (but: too generic?)
- `QUERY ( subquery )` (but: too generic?
- `EVAL ( subquery )` (but: sounds too much like metaprogramming?)
- `$( subquery )` (but: feels like we should reserve this syntax, perhaps for metaprogramming/templating)
- `(| ()-[]->(c) RETURN c |)` - Bananas ftw :)

boggle: Maybe the whole approach is also wrong.. for crpq we're discussing path pattern predicates.. maybe here we're missing inline function or expression-level template definitions? Given those, the quest for scalar subqueries really just becomes a search for anonymous function syntax.

```
DECLARE expr AS subquery(params)
...
WITH expr1 + subquery(...)
...
```

boggle: One more... if we don't allow subquery short forms, i.e. starting with a pattern, round parenthesis might actually suffice from a parsing point of view... `(MATCH (a)-[r]->(b) RETURN r)`. It's inconsistent with the other subquery forms under discussion but it would save us the hassle of coming up with weird fancy bracket syntax.

petraselmer: - `'SCALAR'` as a keyword is simple and "does what it says on the tin"; i.e. it's very clear in its intention. What are the objections it its use?

- `$( subquery )` is rather elegant. Re boggle's comments about the drawback, would it not be reasonable to infer from context?

- The inline function/expression-level template defs option looks very unwieldy from a readability point of view and does not marry up at all well with our other Subquery Family syntax. I don't see any advantage in it.

thobe: Yes, I've had that thought as well...

> if we don't allow subquery short forms /.../ round parenthesis might actually suffice from a parsing point of view

thobe: I don't remember what the objections were to `SCALAR` - perhaps we never suggested it in prior conversation? I know that we suggested `VALUE` and decided that it was _definitely_ too generic.

I'm leaning towards `SCALAR` now, since I agree with Stefan that since we've used `$` for parameters, we should reserve variations with that character for variations of parametrization.

cip/1.accepted/CIP2017-03-29-Single-Value-Subqueries.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+toc::[]
+
+== Scalar Subqueries and List Subqueries
+
+Scalar Subqueries are read-only subqueries that produce a single value in a single row.
+The result of a Scalar Subquery is the single value (in the single row) produced by the subquery.
+
+List Subqueries are read-only subqueries that produce a single value per row, and zero or more rows.
+The result of a List Subquery is to collect the value of all rows produced by the subquery into a list.
```
Mats-SX: This sentence sounds a bit strange to me. Perhaps 

> The goal of a List Subquery

or

> The result of a List Subquery is the list formed by collecting all of the values of all rows produced by the subquery.

cip/1.accepted/CIP2017-03-29-Single-Value-Subqueries.adoc
```diff
@@ -0,0 +1,68 @@
...
+    UNWIND existing_list_of_items AS item
+    WHERE item.size < 15
+    RETURN item
+] AS small_items
+----
+
+[source, cypher]
+.Collect separate lists of friends and enemies
+----
+MATCH (me:Person{name:$my_name})
```
Mats-SX: Cypher style: space before property predicate.
Cypher style: space after colon in property predicate.

cip/1.accepted/CIP2017-03-29-Single-Value-Subqueries.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+ScalarSubquery = 'SCALAR', '(', SingleValueSubquery, ')' ;
+
+SingleValueSubquery = SingleValuePatternSubquery
+                    | SingleValueUnwindSubquery
+                    | SingleValueReadSubquery
+                    ;
+
+SingleValuePatternSubquery = PatternPart, Filter, SingleValueReturn ;
+SingleValueUnwindSubquery = 'UNWIND', Expression, ['AS', Variable, Filter] ;
```
Mats-SX: Would the `UNWIND` form not require a `RETURN` clause?

thobe: The implication here is that the implicitly `RETURN` value is the value unpacked by `UNWIND`.

`UNWIND` with `RETURN` is handled by the regular `SingleValueReadSubquery`.

The use cases for this form is:

1. Unpacking a singleton-list: `SCALAR( UNWIND myList )`
2. Sorting a list: `[UNWIND myList AS x ORDER BY x.age]`
3. Filtering a list: `[UNWIND myList AS x WHERE x.name CONTAINS ' ']`

et.c.

boggle:
Just realized... using round parenthesis for scalar subqueries seems not well aligned with `EXISTS {}` which also returns a single value... in a way.. Not sure how to solve this given that we overloaded `exists(..)`

a-type:
This seems like a very powerful feature and I'd love to see it added. In particular I think this would give a massive boost to the GraphQL->Cypher use case seen the in Neo4j-GraphQL tools. It's been two years without much conversation, but what would it take to get this moving again?

thobe:
@a-type indeed this was first conceived of when we did a comparative analysis of Cypher and GraphQL in 2015. What is primarily needed for this to move forward is an openCypher implementor with a desire to get this into their product. In terms of actual work, the next step in terms of actual work on this proposal is to start designing TCK test cases to guide an implementation.

thobe:
I believe that if we disallowed the `SingleValuePatternQuery` form from `ScalarSubquery` - which has the effect of a `ScalarSubquery` always starting with a keyword - we could do away with the `SCALAR` keyword as a prefix for `ScalarSubquery` with only a single parser lookahead.

I think that would be nicer.

SpaghettiFactory:
Does this enable sorting, filtering and using aggregate functions in Map Projections, as mentioned [in this blog post](https://neo4j.com/blog/cypher-graphql-neo4j-3-1-preview/)? 

Would be awesome if neo4j could be queried like this!

boggle:
Yes, that's the goal of this proposal which I expect to also become a part of GQL. It may happen eventually in Neo4j but note this repository is about language design primarily, not implementation by vendors.

PR #226, opened 2017-04-21
Add namespace to procedure names

Mats-SX:
- Correctly use parentheses for in-query procedure call

### DIFF
--- a/grammar/basic-grammar.xml
+++ b/grammar/basic-grammar.xml
@@ -375,7 +375,7 @@
   </production>
 
   <production name="Namespace" rr:inline="true">
-    <repeat min="1">
+    <repeat>
       <non-terminal ref="SymbolicName"/> .
     </repeat>
   </production>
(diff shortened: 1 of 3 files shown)


grammar/basic-grammar.xml
```diff
@@ -370,9 +370,16 @@
   </production>
 
   <production name="ProcedureName" rr:inline="true">
+    <non-terminal ref="Namespace" rr:title="procedure namespace"/>
     <non-terminal ref="SymbolicName" rr:title="procedure name"/>
   </production>
 
+  <production name="Namespace" rr:inline="true">
+    <repeat min="1">
```
thobe: Whether `min` should be `0` or `1` is still being debated is it not?

Mats-SX: It is.

Mats-SX: @thobe Would you be more comfortable merging this is the `min` is set to `0`?

Mats-SX:
@thobe Changed it so that namespace can be empty.

PR #231, opened 2017-04-24
Categorise reserved words

Mats-SX:
Categorises reserved words into proper groups. Also adds five reserved words for future use: `MANDATORY`, `SCALAR`, `OF`, `ADD`, `DROP`.

### DIFF
--- a/cip/1.accepted/CIP2016-12-19-Reserved-keywords.adoc
+++ b/cip/1.accepted/CIP2016-12-19-Reserved-keywords.adoc
@@ -116,10 +116,15 @@ In addition to this, we list a number of words that are reserved for future use.
 
 ===== Reserved for future use
 
+* ADD
 * CONSTRAINT
 * DO
+* DROP
 * FOR
+* MANDATORY
+* OF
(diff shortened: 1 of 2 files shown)


cip/1.accepted/CIP2016-12-19-Reserved-keywords.adoc
```diff
@@ -53,40 +53,46 @@ By _escaping_ any of the reserved words (encapsulating in backticks ```), they w
...
 * OPTIONAL
-* ORDER
 * REMOVE
 * RETURN
 * SET
+* UNION
+* UNWIND
+* WITH
+
+===== Subclauses
```
petraselmer: I think that the following keywords:

`ASC`, `ASCENDING`, `BY`, `DESC`, `DESCENDING` and `ON` do not fall naturally under "subclauses".

A far more intuitive place would be under a new "Modifiers" section (as these modify some other keyword - most notably, a clause)

Mats-SX:
@petraselmer Please see updated content.

Issue #221, opened 2017-04-19
Unobservable behavior in TCK test


zsambek:
As a university project (under the supervision of @szarnyasg and @jmarton), I tried to implement the TCK tests on Neo4j. My first approach was to use Cypher queries to get the side-effects, but this turned out to be unfeasible, e.g. `CREATE (n) DELETE n` does not result in any observable behaviour. I then switched to using the [`QueryStatistics`](http://neo4j.com/docs/java-reference/current/javadocs/org/neo4j/graphdb/QueryStatistics.html) class, which works fine for nodes using the `getNodesCreated()` and `getNodesDeleted()` methods. However, it does not provide method for retrieving properties.

```
  Scenario: Non-existent values in a property map are removed with SET =
    Given any graph
    And having executed:
      """
      CREATE (:X {foo: 'A', bar: 'B'})
      """
    When executing query:
      """
      MATCH (n:X {foo: 'A'})
      SET n = {foo: 'B', baz: 'C'}
      RETURN n
      """
    Then the result should be:
      | n                         |
      | (:X {foo: 'B', baz: 'C'}) |
    And the side effects should be:
      | +properties | 2 |
      | -properties | 1 |
```

szarnyasg:
So, basically, our question is twofold:
1. what is the recommended approach for testing such behaviour?
2. what is the precise semantics of the `+properties` and `-properties` side effects?

For question 1, I believe we should only test side effect that can be observed with running standard openCypher queries. For example, I think if a user runs the query `CREATE (n) DELETE n` and the database does nothing, that classifies as a correct behaviour. Even if it does create and delete a node, the user has no way of telling whether this happened or not (except for extracting internal ids and making assumptions based on their values, but that is bad practice in general).

For question 2, I do not have an answer that is in sync with the [test case](https://github.com/opencypher/openCypher/blob/72603db1277038c681797abc8d163397166f7ba7/tck/features/SetAcceptance.feature#L269):
* If `properties` means _property keys_, then it should be `+1`, `-1`, as the `bar` key is dropped, while the `baz` key is added (and the `foo` key is kept).
* If `properties` means _property key-value pairs_, then it should be `+2`, `-2`, as the `foo: 'A'` and `bar: 'B'` are dropped, while `foo: 'B'` and `baz: 'C'` are added.
* If `properties` means _property values_, then it should be `+1`, `-1`, as the `'A'` is dropped, while `'C'` is added (and `'B'` is kept).

[...]

PR #236, opened 2017-05-15
Clearly define the meaning of TCK side effects

Mats-SX:
At the 2nd oCIM, @szarnyasg presented the problem of unclearly defined side effects in the TCK, which is also discussed in #221. @szarnyasg also proposed a solution, which is largely implemented in this PR, with one notable amendment: properties are defined not only by their key and value, but also by the containing entity.

This PR aims to solve #221.

### DIFF
--- a/tck/README.adoc
+++ b/tck/README.adoc
@@ -102,6 +102,63 @@ If the side effects step reads `And no side effects`, this implies that all quan
 
 For 'negative' tests, where errors are expected (see <<errors>>), it is implied that the graph suffers no side effects.
 
+==== Observability of side effects
+
+In order for a side effect to be reported, it has to be _observable_ from the point of view of a subsequent Cypher query executed against the same graph.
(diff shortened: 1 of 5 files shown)


tck/README.adoc
```diff
@@ -102,6 +102,63 @@ If the side effects step reads `And no side effects`, this implies that all quan
...
+RETURN n AS entity, key, properties[key] AS value
+UNION ALL
+MATCH ()-[r]->()
+UNWIND keys(r) AS key
+WITH properties(r) AS properties, key, r
+RETURN r AS entity, key, properties[key] AS value
+----
+
+Note that in the definition above, a property is defined as the triple of containing entity, key, and value.
+Therefore the operation of copying a property between two entities will be noted as one removal and one addition in the side effects.
```
szarnyasg: We believe _copying_ a property between two entities should be noted as one addition, while _moving_ should be noted as one removal and one addition.

Mats-SX: Agreed.

PR #241, opened 2017-07-02
CIP2017-06-18 Multiple Graphs

boggle:
This is an *early draft* for a proposal for making Cypher work with multiple graphs


cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+
+This section introduces the *multiple graphs execution model* as an evolution of the *single graph execution model* that enables the addition of features to the language for working with multiple graphs, i.e. it changes the basis of graph matching and graph manipulation.
+
+This CIP proposes the adoption of the multiple graphs execution model by Cypher and to execute existing, single graph queries under this model as outlined below.
+
+=== Model definition
+
+This CIP proposes to redefine the *execution context* to be
+
+* a set of graphs in multiple graphs execution model
```
Mats-SX: missing a `the`

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+This CIP proposes to redefine the *query context* to be
+
+* a set of named graphs from the *execution context*
+* an optional information that indicates which of these named graphs if any is the *source graph*
+* an optional information that indicates which of these named graphs if any is the *target graph*
+* optional *tabular data*, i.e. a potentially ordered bag of records, each having the same fixed set of fields
+
+These redefinitions constitute the multiple graphs execution model. A parameterized Cypher query under this model can _also_ be described as executing within (and operating on) a given execution context and an initial query context and finally returning the query context produced as output for the top-most `RETURN` clause.
+
+As a consequence of adopting the new the multiple graphs execution model, the semantics of each clause needs to be (re-)define how the execution of the clause transforms all given input query contexts into an output query context
```
Mats-SX: A `the` too many

Mats-SX: (re-)define`d` as to how ...

Mats-SX: preserve~~s~~

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+
+These rules ensures that a follow-up query (or a clause) that is executed on a partial query context is always provided with a source graph for graph matching, a target graph for graph manipulation, and tabular data such that it will iterate over at least one row.
+
+=== Query composition
+
+The multiple graphs execution model provides a natural way for the sequential composition of queries:
+
+A query `Q1` whose output signature is an acceptable (in terms of provided bindings) input signature for another query `Q2` may be composed sequentially with `Q2` into a new query `Q3` that first runs `Q1` on the initial query context, next runs `Q2` on the query context returned by `Q1`, and finally returns the query context returned by `Q2`.
+
+This homogenous query composition is enabled by using an uniform query context that is passed between clauses.
```
Mats-SX: `an` -> `a`

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+
+Note: The currently drafted subquery CIP proposes a language addition (e.g. `THEN`) for expressing this kind of query composition directly. In terms of this CIP, `THEN` is simply syntactic sugar for `WITH * GRAPHS *`
```
Mats-SX: This should link to the referenced CIP, as several subquery CIPs exist.

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+
+=== Query combinators
+
+Query combinators only need to handle tabular query contexts in the single graph execution model.
+
+For the multiple property graphs execution model, it is necessary to define how query combinators combine the query contexts of all child queries into a new result query context (c.f. query composition).
+
+This CIP proposes that `UNION` (and `UNION ALL` respectively) compute the union between both the source and target graphs, as well as the tabular data (i.e. the records) of all child queries to combine their query contexts.
```
Mats-SX: So this means that `q1 UNION q2` followed by `q3` (in a composition) will let `q3` match from the union of `q1`'s and `q2`'s source graphs? And for updating statements, if the target graph(s) of `q1` and `q2` are different, updates made by `q3` will actually no be reflected in those target graphs?

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+INTO < graph-specifier >
+----
+
+=== FROM clause
+
+The newly introduced `FROM` clause may be used to change both the source and the target graph of the current query context to the graph described by the given `<graph-specifier>`.
+
+=== INTO clause
+
+The newly introduced `INTO` clause may be used to only change the target graph of the current query context to the graph described by the given `<graph-specifier>`.
```
Mats-SX: I would remove `only`

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+This CIP defines the notion of `<graph-specifier>` as a means to introduce, refer to, and alias named graphs.
+
+This CIP proposes the following kinds of graph specifiers:
+
+* `NEW GRAPH [<new-graph-name>] [AT <graph-url>]`: Reference to a newly created, empty graph that is to be bound as `<new-graph-name>` and may potentially overwrite any pre-existing graph at the provided `<graph-url>`
+* `GRAPH [<new-graph-name] AT <graph-url>`: Reference to the graph at the given `<graph-url>` that is to be bound as `<new-graph-name>`
+* `[GRAPH] <graph-name> [AS <new-graph-name>]`: Reference to an already bound named graph
+* `COPY [GRAPH] <graph-name> [AS <new-graph-name>]`: Reference to a copy of an already bound named graph
+* `SOURCE GRAPH [<new-graph-name>]`: Reference to the currently _provided source graph_, optionally to be bound as `<new-graph-name>`
+* `TARGET GRAPH [<new-graph-name>]`: Reference to the currently _provided target graph_, optionally to be bound as `<new-graph-name>`
```
Mats-SX: Formatting suggestion: use second-level bullet points for the prose definitions

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+
+If a graph specifier is not referencing an already bound named graph and does not specify a `<new-graph-name>`, it is bound to a fresh system generated name.
+The details of this are left to implementations.
+
+It is an error to use a `<graph-specifier>` in a context where it's introduced `<new-graph-name>` is already bound.
```
Mats-SX: `it's` -> `its`

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+
+=== Returning, aliasing, and selecting graphs
+
+This CIP proposes to extend both the `WITH` and the `RETURN` clauses with new syntax for controlling the set of available named graphs that should be passed on by the clause (or returned from the query respectively) by explicitly specifying all `<graph-return-items>`.
+The newly proposed syntax is:
+
+[source, cypher]
+----
+WITH [ < return-items > ] [ [ INPUT ] GRAPHS < graph-return-items > ]
+RETURN [ < return-items > ] [ [ INPUT ] GRAPHS < graph-return-items > ]
```
Mats-SX: `<return-items>` are not optional in `WITH` and `RETURN`. There has to be at least one.

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+----
+
+This CIP proposes the following kinds of `<graph-return-items>`:
+
+* `<graph-specifier-list>`: A comma separated list of `<graph-specifier>` that are to be passed on
+* `*`: All named graphs are to be passed on
+* `*, <graph-specifier-list>`: All named graphs are to be passed on together with any additional named graphs that are newly bound in `<graph-specifier-list>`
+* `-`: No named graphs are to be passed on
```
Mats-SX: I'm interpreting that `GRAPHS` is optional (which I support). What is the point of `GRAPHS -` if we can just leave it out?

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+
+The order of named graphs inherently given by `<graph-return-items>` is semantically insignificant.
+However it is recommended that conforming implementations preserve this order at least in programmatic output operations (e.g. a textual display of the list of returned graphs).
+This in essence mirrors the semantics for tabular data returned by Cypher.
+
+Both `WITH ... GRAPHS ...` and `RETURN ... GRAPHS ...` will pass on (or return respectively) exactly the set of described named graphs.
+To simplify passing on available graphs it is proposed by this CIP that regular `WITH <return-items>` is taken to be syntactic sugar for `WITH <return-items> GRAPHS -` and that regular `RETURN <return-items>` is taken to be syntactic sugar for `RETURN <return-items> GRAPHS -`.
```
Mats-SX: What is the point of having a long-form `GRAPHS -` as the normal form, and call leaving it out syntactic sugar? Why not say that leaving it out is the normal form, and that the other forms modify that?

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+
+To even further simplify, it is additionally proposed that `WITH|RETURN <return-items> INPUT GRAPHS <graph-return-items>` is to be syntactic sugar for `WITH|RETURN <return-items> GRAPHS <graph-return-items>, SOURCE GRAPH, TARGET GRAPH`.
```
Mats-SX: I'm not convinced of the usefulness of this syntactic sugar -- I find that it is hard to know what kind of queries will be prominent in this new model. In general, I think that it would be useful to have a little less focus on the syntactic sugar bits, and more on the core model. Syntactic sugar additions could always follow later.

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+However if `<graph-return-items>` already passes on a reference for the `SOURCE GRAPH`, no additional reference for it is added and if `<graph-return-items>` already passes on a reference for the `TARGET GRAPH`, no additional reference for it is added.
+
+If the current named source graph (or the current named target graph) are not passed on, they are discarded and due to the rules regarding partial query contexts the provided source graph (or target respectively) again are chosen to be the default graph of the outer execution context.
+
+Note: `WITH <return-items> GRAPHS *` may be used to pass through the initial query context without having to alias incoming source and target graphs explicitly.
+
+=== Discarding available tabular data
+
+It is additionally proposed that both `WITH GRAPHS <graph-return-items>` and `RETURN GRAPHS <graph-return-items>` are syntactic sugar for `WITH - GRAPHS <graph-return-items>` (and `RETURN - GRAPHS <graph-return-items>` respectively).
```
Mats-SX: I feel similarly to this as to `GRAPHS -`; I prefer the absence of `-` to its presence in this context.

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+RETURN count(p) AS num_triangles GRAPHS swedish_triangles, sweden_people, german_people
+----
+
+== Interaction with existing features
+
+This proposal is far reaching as it changes both the property graph model and the execution model of the language.
+
+However, the change has been carefully designed to not change the semantics of existing queries.
+
+== Alternatives
```
Mats-SX: I think this and subsequent sections are superfluous since the introduction of CIRs. We should modify our template.

cip/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,476 @@
...
+This CIP defines an instance of the *multiple property graphs model* to be a set of property graphs that adhere to the definitions given below.
+
+=== (Property) Graph
+
+_Definition_ A *property graph* is a set of labeled nodes and typed relationships both together with their properties (a property is a tuple of a named key and a value).
+Graphs may be updatable, i.e. the set of contained nodes and relationships may change during the lifetime of the graph.
+It is an error to attempt to update a read-only graph.
+
+The same node or relationship may be part of many graphs.
+A relationship may only be part of a graph if it's start node and it's end node are both also part of the same graph.
```
boggle: That always trips me up :)

cip/CIP2017-06-18-multiple-graphs.adoc
boggle: The procedures CIP (?) I think added `-` for procedures not returing any columns, similarly researches have suggested, that's an omission on the part of SQL to not be able to return no columns (more so for Cypher where the single row field less table plays a special role to start off queries). In light of this I added this for no reason but consistency with these other decisions.

cip/CIP2017-06-18-multiple-graphs.adoc
boggle: Let's revisit how default graphs are handled as a group first - this may very well remove the need for this. In short I added this as a simple way for a query to say: "I'm ok to run on any incoming graphs and am happy to pass those on, just give 'em some names for me". Without this sugar, expressing this becomes rather verbose.

cip/CIP2017-06-18-multiple-graphs.adoc
Mats-SX: Yeah, used to trip me up too, but then I learned that `it's == it is`, so in case you're unsure, just spell it out and it'll become apparent :)

cip/CIP2017-06-18-multiple-graphs.adoc
Mats-SX: But for procedures there was actually a need for `YIELD -` in order to not cause implicit conflicts with variables that were in scope, as I recall. My personal preference would be to use the empty string to denote the intention in this proposal.

cip/CIP2017-06-18-multiple-graphs.adoc
Mats-SX: I'm not against the sugar per se, I just find it difficult to assess whether a particular piece of sugar is valuable this early in the process of defining these very new concepts, and so I'm leaning towards skepticism in general. I find it is peripheral to the contents of the CIP anyway.

Mats-SX:
Great work putting these concepts into spec!

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+
+This CIP proposes the following update semantics for Cypher with support for multiple graphs:
+
+* All updating clauses read from the source graph and write to the target graph of their current query context.
+  More concretely:
+  ** Entities are always created in, updated in, and deleted from the currently provided target graph.
+  ** All entities of bound pattern variables in `CREATE` and `MERGE` are always added to the provided target graph of the current query context.
+  ** `MERGE` still uses the source graph to find existing entities
+  ** Deleting an entity only affects the provided target graph of the current query context.
+* Updating queries always return all variables and graphs in scope, i.e. the behave as if they would end in `RETURN *, GRAPHS *` (This syntax is introduced below).
```
Mats-SX: i.e. the`y` behave

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+* Semantically, all effects of an updating clause must be made visible before proceeding with the execution of the next clause.
+In other words, a conforming implementation must ensure that a later clause always sees the complete set of updates of a preceding updating clause.
+
+This CIP proposes allowing `MERGE` to be followed by a non-empty, comma-separated list of bound variables for explicitly adding an entity to the target graph.
```
Mats-SX: an entity -> entities

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
Mats-SX: Isn't this a massive change? Currently, updating queries return nothing at all.

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+
+=== Simulating the single graph execution model
+
+Execution under the single graph execution model can be simulated in the multiple graphs execution model by executing the query in an session context that uses the single graph as the default graph, and by running it on an empty initial query context.
```
Mats-SX: What about queries that end with an updating clause? They will return vs not return in the two models, no?

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
+= CIP2017-06-18 - Multiple Graphs
+:numbered:
+:toc:
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Author:* Stefan Plantikow <stefan.plantikow@neotechnology.com>
+
+This material is based on internal contributions of Alastair Green <alastair.green@neotechnology.com>, Mats Rydberg <mats.rydberg@neotechnology.com>, Martin Junghanns <martin.junghanns@neotechnology.com>, Tobias Lindaaker <tobias.lindaaker@neotechnology.com>.
```
boggle: Fixed.

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+
+[abstract]
+.Abstract
+--
+This CIP proposes extending Cypher to support the construction, transformation, and querying of multiple graphs (i) by adopting the proposed multiple property graphs model, (ii) by adopting a new execution model, namely the 'multiple graphs execution model', and (iii) by introducing new syntax for working with multiple graphs.
```
boggle: Fixed.

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+
+A Cypher query currently takes a set of parameters as input, queries a single, global, implicit property graph, potentially updates it, and finally returns tabular data derived from it.
+Query parameters are conceptually thought to be inlined before the start of query execution.
+Therefore the *session context* of a whole Cypher query is a single, global, implicit property graph.
+
+Each sequence of clauses (sometimes called a *pipeline*) optionally operates on this single implicit graph and takes a single table input in order to produce a new single table output.
+Furthermore, Cypher supports query combinator clauses like `UNION` and `UNION ALL` for merging two pipelines into a single pipeline.
+Therefore the *query context* that conceptually is passed between clauses in the single graph execution model is simply a single table.
+
+With this terminology in place, execution of a parameterized Cypher query in the single graph execution model can be described as executing within (and operating on) a given session context and an initial query context and finally returning the query context produced as output for the top-most `RETURN` clause.
```
boggle: I'm thinking in planner trees :) Fixed.

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+
+Note: This formulation is introduced to describe a high-level model for the execution of queries; a real world implementation is free to choose any other internal representation (e.g. based on an algebra) as long as it does not violate the specified semantics.
+
+== Proposal: Multiple Graphs Execution Model
+
+In the single graph execution model, tabular data serves as the basis of iteration while the single implicit global graph serves as the basis of graph matching and graph manipulation.
+
+This section introduces the *multiple graphs execution model* as an evolution of the *single graph execution model* that enables the addition of features to the language for working with multiple graphs, i.e. it changes the basis of graph matching and graph manipulation.
```
boggle: Fixed.

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+
+This CIP proposes redefining the *query context* to be
+
+* a set of named graphs from the *session context*
+* *tabular data*, i.e. a potentially ordered bag of records, each having the same fixed set of fields
+* a special graph drawn from the *session context* that is called the *source graph*
+* a special graph drawn from the *session context* that is called the *target graph*
+
+These redefinitions comprise the multiple graphs execution model.
+A parameterized Cypher query under this model can _also_ be described as executing within (and operating on) a given session context and starting from an initial query context and finally returning the query context produced as output for the top-most `RETURN` clause.
```
boggle: Fixed.

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+Therefore the *session context* of a whole Cypher query is a single, global, implicit property graph.
+
+Each sequence of clauses (sometimes called a *pipeline*) optionally operates on this single implicit graph and takes a single table input in order to produce a new single table output.
+Furthermore, Cypher supports query combinator clauses like `UNION` and `UNION ALL` for merging two pipelines into a single pipeline.
+Therefore the *query context* that conceptually is passed between clauses in the single graph execution model is simply a single table.
+
+With this terminology in place, execution of a parameterized Cypher query in the single graph execution model can be described as executing within (and operating on) a given session context and an initial query context and finally returning the query context produced as output for the top-most `RETURN` clause.
+
+Note: This formulation is introduced to describe a high-level model for the execution of queries; a real world implementation is free to choose any other internal representation (e.g. based on an algebra) as long as it does not violate the specified semantics.
```
boggle: Fixed.

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+
+[source, cypher]
+----
+FROM < graph-spec >
+INTO < graph-spec >
+----
+
+==== FROM clause: Change the source and the target graph
+
+This CIP proposes a new `FROM` clause to change both the source and the target graph of the current query context as described.
```
Mats-SX: I'm not convinced that having `FROM` change _both_ source and target is the best choice, but I'm also not sure how to argue in either direction. It feels easier to grasp if `FROM` is tied to source and `INTO` tied to target.

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
```diff
@@ -0,0 +1,944 @@
...
+
+`FROM DEFAULT GRAPH` may be used to discard the current source and the current target graph.
+
+==== INTO clause: Change the target graph only
+
+This CIP proposes a new `INTO` clause to change the target graph of the current query context as described.
+
+`INTO <graph-def>` aliases the defined graph to the given `<new-graph-name>`.
+
+`INTO DEFAULT GRAPH` may be used to discard the current source and the current target graph.
```
Mats-SX: This is a copy-paste error, right? `INTO` would only discard the current target graph, I assume.

cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc
Mats-SX: This comment is now obsolete.

linsimiao:
hi all, I have read your [documentation](https://github.com/boggle/openCypher/blob/CIP2017-06-18-multiple-graphs/cip/1.accepted/CIP2017-06-18-multiple-graphs.adoc ). I found it easy to mix `CONSTRUCT` with `UPDATE`. I wonder whether the following cyphers mean the same.

`FROM  xxx
match (a:Person)
UPDATE GRAPH
merge (b:Student{name:a.name})`

and


`FROM xxx
match (a:Person)
CONSTRUCT 
merge (b:Student{name:a.name})`

suppose the working graph is yyy, is it both two cyphers will lead to create nodes with a label `Student` in the graph yyy. 

Thanks for your reply.

PR #244, opened 2017-07-19
Expands List comparability definition

alexaverbuch:


petraselmer:
As this is an accepted CIP, we need to ascertain whether or not this change is simply propagated as-is, or whether it needs to be part of an "Addendum" section. 

For this reason, I have added the `NOT READY FOR MERGE` label

sherfert:
Merged into #262

PR #250, opened 2017-08-29
Align side effects with standard definitions

Mats-SX:
This updates the side effects specification of many TCK scenarios, following the definitions from #236.

It is worth to note that the standard definition of the `labels` side effect metric measures the amount of distinct labels in the graph, and not the amount of nodes carrying these labels. This has the consequence that some queries that set or remove labels from nodes will not report any side effects at all, under certain circumstances. For example, this is a valid scenario:

```
Scenario: Setting an already existing label does not constitute a side effect
  Given any graph
  And having executed:
    """
    CREATE (:A), (:B)
    """
  When executing query:
    """
    MATCH (b:B)
    SET b:A
    """
  Then the result should be empty
  And no side effects
```

It is beyond the scope of this PR to change this, but it could be a topic for the oCIG to consider.

### DIFF
--- a/tck/README.adoc
+++ b/tck/README.adoc
@@ -148,6 +148,7 @@ RETURN r AS entity, key, properties[key] AS value
 
 Note that in the definition above, a property is defined as the triple of containing entity, key, and value.
 Therefore the operation of moving a property from one entity to another will be noted as one removal and one addition in the side effects.
+Likewise, the operation of changing the value of a property is noted as one removal and one addition in the side effects.
 
 ===== Labels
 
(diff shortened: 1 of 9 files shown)


Mats-SX:
@szarnyasg This might interest you.

szarnyasg:
@Mats-SX seems good to me!

The example I used in [OCIM2](https://s3.amazonaws.com/artifacts.opencypher.org/website/ocim2/slides/ocim2-tck.pdf#page=24) is now `+properties 2` and `-properties 2`, which is more intuitive (and easier to argue for): https://github.com/opencypher/openCypher/pull/250/commits/aee23f4d61647ad789e192460ec2d36023992902#diff-1e889763b10d13e1a0835e213c84e218R272

Interestingly, it seems `-labels` is a "new thing", i.e. it was not present before.

szarnyasg:
> -labels is not all new, though it may have received increased use in this PR (which also makes changes to some scenarios that were incorrect even under the old, informal side effect definitions). You can find it here for example.

Good point. It seems [ag](https://geoff.greer.fm/ag/)'s literal search tricked me: `ag --literal "\-labels"` showed no result for some reason. Nice to know :-).

PR #261, opened 2017-10-17
Specify Cypher 9

Mats-SX:
Outlines a proposal for Cypher 9 -- the first official standard Cypher version as governed by the openCypher project.

The specification is made up of three resources: the Neo4j 3.2 Manual, the Standardisation Scope document, and this CIP. Together these three components define Cypher 9.


cip/1.accepted/CIP2017-10-17-Cypher-9.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+* `any()`
+* `all()`
+* `none()`
+* `single()`
+* `reduce()`
+
+New syntax for these constructs needs to be proposed to consider them applicable for the standard.
+
+The `exists()` function is defined in an accepted CIP, and does not need to rely on vendor documentation.
```
petraselmer: Aren't we including the form `exists(n.prop)` and excluding the form `exists(pattern)`? This sentence is not very clear.

docs/standardisation-scope.adoc
```diff
@@ -200,20 +224,17 @@ It is the goal of this project to create a good and feature-rich standard langua
 === Functions
 
 * `all()`
-* `allShortestPaths()`
 * `any()`
 * `distance()`
+* `exists()`
```
petraselmer: Please see my earlier comment about the two forms of `exists()`

cip/1.accepted/CIP2017-10-17-Cypher-9.adoc
Mats-SX: Not as documented in the Neo4j manual, but as specified in the CIP. I thought a CIP is better than referring to the manual?

Mats-SX: My interpretation on the Standardisation Scope is that it only acts as an overlay of what's in Neo4j ('s manual), and doesn't reference anything else.

docs/standardisation-scope.adoc
Mats-SX: It's not included as implemented in Neo4j, but it is included as specified in the EXISTS CIP.

Mats-SX:
@petraselmer Do you think this is obsoleted in the presence of the Cypher 9 reference manual on opencypher.org?

petraselmer:
@Mats-SX I must say I have found this very useful. At a glance, one can see very easily what's in and what's "out". So, I would say it is well worth having this as a 'quick reference' type of resource.

Mats-SX:
Then I suggest we merge it :)

PR #262, opened 2017-10-30
Clarification about null in lists in CIP2016-06-14

Lojjs:
The Neo4j Cypher team recently fixed that our implementation didn't follow the 2016-06-14 CIP regarding comparisons and equality with lists containing null. However, during the work we found the CIP more confusing than helpful so I have tried to clarify the relevant parts of it and increase the number of examples.

### DIFF
--- a/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
+++ b/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
@@ -71,6 +71,53 @@ For the following discussion, it is helpful to clarify the meaning of `null`. In
 [[unknown-null,unknown `null`]]*Unknown*:: An "unknown" `null` is taken to be a placeholder for an arbitrary but unknown value. When evaluating predicates, an "unknown" `null` is the `maybe` truth value of ternary logic. For node and relationship properties, an "unknown" `null` is a value that is definite in the real world but has not been stored in the graph. Since in these cases, two "unknown" `null` values stand for arbitrary but definite values in the real world, two "unknown" `null` values should never be treated as certainly being the same value.
 [[missing-null,missing `null`]]*Missing*:: A "missing" `null` is taken to be a marker for the absence of a value. In the context of updating node properties from a map, a "missing" `null` is used to mark properties that are to be removed. In the context of `DISTINCT` and grouping, a "missing" `null` value is used as grouping key for all records that miss a more specific value. Since in these cases, two "missing" `null` values represent the same concept, they should always be treated as the same value.
 
+===== Ternary logic truth tables
+
+Allowing null for an unknown boolean value requires defining ternary logic for boolean operators.
+The truth tables for the boolean operators `NOT`, `AND` and `OR` are given below.
+The `XOR` in Cypher is defined by these basic operators as `a XOR b = (a AND NOT(b)) OR (NOT(a) AND b)`.
+
+*Negation*
+[width="50%",options="header"]
+|===========
+|a    |NOT a
+|true |false
+|null |null
+|false|true
+|===========
+
+
+*Conjunction*
+[width="50%",options="header"]
+|===================
+|a    |b    |a AND B
+|true |true |true
+|true |null |null
+|true |false|false
+|null |true |null
+|null |null |null
+|null |false|false
+|false|true |false
+|false|null |false
+|false|false|false
+|===================
+
+
+*Disjunction*
+[width="50%",options="header"]
+|==================
+|a    |b    |a OR B
+|true |true |true
+|true |null |true
+|true |false|true
+|null |true |true
+|null |null |null
+|null |false|null
+|false|true |true
+|false|null |null
+|false|false|false
+|==================
+
 [[regular-map]]
 ==== Regular maps
 
--- a/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
+++ b/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
@@ -120,7 +120,7 @@ We propose that <<comparability-con,comparability>> should be defined between an
   * Lists are compared in dictionary order, i.e. list elements are compared pairwise in ascending order from the start of the list to the end. Elements missing in a shorter list are considered to be less than any other value (including `null` values).
   For example, `[1] < [1, 0]` but also `[1] < [1, null]`.
   * If comparing two lists requires comparing at least a single `null` value to some other value, these lists are <<incomparable>>.
-  For example, `[1, 2] >= [1, null]` evaluates to `null` (ncomparable), but `[1, 2] >= [3, null]` evaluates to `false`.
+  For example, `[1, 2] >= [1, null]` evaluates to `null` (incomparable), but `[1, 2] >= [3, null]` evaluates to `false`.
   * Lists are <<incomparable>> to any value that is not also a list.
 - Maps
   * [[regular-maps,regular maps]]Regular maps
--- a/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
+++ b/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
@@ -130,8 +130,8 @@ Cypher today has one supertype `MAP` for all map values. This includes nodes (of
 We propose that comparability and equality should be aligned with each other, i.e.
 
 
-`expr1 = expr2` = `expr1 >= expr2 AND expr1 \<= expr2`.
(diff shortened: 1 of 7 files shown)


cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -164,13 +164,20 @@ In order to align equality with <<comparability-def>>, we change equality of lis
...
 
-* `l1` and `l2` must have the same size, i.e. inversely `size(l1) <> size(l2) \=> l1 <> l2`
+* `l1` and `l2` must have the same size, i.e. inversely `size(l1) <> size(l2) \=> l1 <> l2`. This also applies when any of the lists contains null, i.e. `[a1, a2, ..., an] = [b1, b2, ..., null, ..., bm]` evaluates to false if `m <> n`.
 * the pairwise elements of both `l1` and `l2` must be equal, i.e.
 ----
 [a1, a2, ..., an] = [b1, b2, ..., bn]
 <=>
 a1 = b1 && a2 = b2 && ... && an = bn
 ----
+* If the second test requires comparing at least a single `null` value to some other value, these lists are <<incomparable>>. E.g., `[1, 2] = [1, null]` evaluates to `null` (incomparable), but `[1, 2] = [null, 'foo']` evaluates to `false`.
```
Mats-SX: This is still slightly confusing, I think. The second example _does_ compare a `null` value; it is just eaten up by the logical `AND` since `null AND false := false`.

Perhaps we could formulate like so:

> If the second test evaluates to `null`, these lists are <<incomparable>>. This will only happen if at least one element in at least one list is `null` while all non-null elements are equal. For example, `[1, 2] = [1, null]` evaluates to `null` (incomparable), while `[1, 2] = [null, 'foo']` evaluates to `false`.

Lojjs: @Mats-SX Just found this comment , had completely forgot about this PR. But I like your formulation, please change to that wording

cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -164,13 +164,20 @@ In order to align equality with <<comparability-def>>, we change equality of lis
 Specifically, we propose to redefine how equality works for lists in Cypher.
 To determine if two lists `l1` and `l2` are equal, we propose two simple tests, as exemplified by the following:
 
-* `l1` and `l2` must have the same size, i.e. inversely `size(l1) <> size(l2) \=> l1 <> l2`
+* `l1` and `l2` must have the same size, i.e. inversely `size(l1) <> size(l2) \=> l1 <> l2`. This also applies when any of the lists contains null, i.e. `[a1, a2, ..., an] = [b1, b2, ..., null, ..., bm]` evaluates to false if `m <> n`.
```
Mats-SX: Each sentence on its own line.
```suggestion
* `l1` and `l2` must have the same size, i.e. inversely `size(l1) <> size(l2) \=> l1 <> l2`. 
  This also applies when any of the lists contains null, i.e. `[a1, a2, ..., an] = [b1, b2, ..., null, ..., bm]` evaluates to false if `m <> n`.
```

cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -164,13 +164,20 @@ In order to align equality with <<comparability-def>>, we change equality of lis
...
 * the pairwise elements of both `l1` and `l2` must be equal, i.e.
 ----
 [a1, a2, ..., an] = [b1, b2, ..., bn]
 <=>
 a1 = b1 && a2 = b2 && ... && an = bn
 ----
+* If the second test evaluates to `null`, these lists are <<incomparable>>. This will only happen if at least one element in one of the lists is `null` while all non-null elements are pairwise equal. For example, `[1, 2] = [1, null]` evaluates to `null` (incomparable), while `[1, 2] = [null, 'foo']` evaluates to `false`.
```
Mats-SX: This contradicts the above example where `[1, 2] >= [1, 'text']` is said to be `null`, which proves that incomparability between lists is not at all dependent on an existence of `null` values within the lists.

Mats-SX: Also, why would `[1, 2] = [null, 'foo']` become `false`? It's `1 = null AND 2 = 'foo'` which is `null AND null` which is `null`, isn't it?

cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -164,13 +164,20 @@ In order to align equality with <<comparability-def>>, we change equality of lis
...
+* For nested lists, the two tests above are carried out recursively, i.e.
+---
+[[1]] = [[1], [null]] evaluates to false
```
Mats-SX: Because sizes are different, OK! Let's add a comment saying that, I think.

cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -164,13 +164,20 @@ In order to align equality with <<comparability-def>>, we change equality of lis
...
+[[1, 2], [1, 3]] = [[1, 2], [null, 'foo']] evaluates to false
```
Mats-SX: Should be `null` because `1 = null AND 3 = 'foo'` is `null`, right?

cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -157,19 +209,39 @@ We propose that <<comparability-con,comparability>> should be defined between an
...
+<=> size(["a"]) = size[1]) AND "a" = 1
+<=> size(["a"]) = size[1]) AND "a" = 1
+<=> true                   AND null
+<=> null
+----
+
+The same logic applies recursively when comparing nested lists.
+----
+    [[1]] = [[1], [null]]
+<=> size([[1]]) = size([[1], [null]]) AND [1] = [1] AND null = [null]
```
Mats-SX: Just for the sake of the example, we could explain this with an additional step I think:
```suggestion
<=> size([[1]]) = size([[1], [null]]) AND [1] = [1]                       AND null = [null]
<=> 1 = 2                             AND size([1]) = size([1]) AND 1 = 1 AND null
```

cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc
```diff
@@ -157,19 +209,39 @@ We propose that <<comparability-con,comparability>> should be defined between an
...
+<=> false                             AND true      AND null
```
Mats-SX: If above suggest is applied, also apply this
```suggestion
<=> false                             AND true                 AND true AND null
``` size([1]) = size([1]) AND 1 = 1

File: tools/tck-api/src/test/scala/org/opencypher/tools/tck/CypherValueParserTest.scala [hunk 8]
```diff
@@ -62,6 +62,7 @@ class CypherValueParserTest extends FunSuite with Matchers {
     CypherValue("''") should equal(CypherString(""))
     CypherValue("'-1'") should equal(CypherString("-1"))
     CypherValue("null") should equal(CypherNull)
+    CypherValue("NaN") should equal(CypherNaN)
```
Mats-SX: @sherfert This is interesting. Are you intending to write TCK scenarios that assert `NaN`s being returned from Cypher? The language doesn't have a literal syntax for it at the moment (although you could mimic it with e.g. `1.0/0.0`). I believe it has been said in the past that Cypher should convert `NaN`s to `null` as a result value, to unify the values.

sherfert: I don't perceive the lack of a literal as problematic for having it in result assertions in TCK tests.

I actually did write these tests in ef32837 . If the expectation is that Cypher converts them to `null`, that is not what neo4j is doing at least. 

Sorry for not waiting on another review and merging pre-maturely. What do you think how we should proceed with `NaN`s?

Mats-SX: 1. I agree, tentatively.

2. Right.

3. I think we should discuss it in the CLG.

PR #278, opened 2018-01-03
Fixed broken link

KingOfThePirates:
One concern with this pull request: The title (CIP2015-06-16..) doesn't match the document name (CIP2015-09-16..). I searched a bit but this seems to be the only result

### DIFF
--- a/tck/README.adoc
+++ b/tck/README.adoc
@@ -178,7 +178,7 @@ Note that this definition measures the amount of distinct labels present in the 
 === Format of the expected results
 
 Values that can be returned from Cypher can be categorized into three groups: primitives (integers, floats, booleans, and strings), containers (lists and maps), and graph elements (nodes, relationships, and paths).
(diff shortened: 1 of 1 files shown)


Mats-SX:
Hello @KingOfThePirates and good catch. I've pushed a rename of the title to match the name of the file; the latter is correct, which can be verified by inspecting the history of the file (it was 09 originally and changed to 06 at one point).

If you can rebase your PR onto another recent fix for the license headers we can get a green build and merge this PR.

KingOfThePirates:
Hi @Mats-SX, I apologize for the noobiness, but how might I accomplish this correctly? I read a bit about git rebase but I'm still confused. And I don't know how to find a recent fix for the license headers - or were you wanting me to make a PR for that as well and squash them?

Mats-SX:
@KingOfThePirates It's fine, I merged it anyway as it's just a document test change. Thanks for your contribution.

For the record, I was referring to something like this:

```
// in main folder of your PR branch locally
$ git fetch origin // assumes you've bound this repository to the origin remote
$ git rebase origin/master
$ git push -f // needs force because you've put your commit on top of a different history
```

PR #281, opened 2018-01-04
Fix side effect

SaschaPeukert:
Builds on top of 1c8fda6

Either we should calculate side effects more precisely or only make the assertion on an empty graph. We opted for making the assertion on an empty graph as the easier short-term solution

### DIFF
--- a/tck/features/CreateAcceptance.feature
+++ b/tck/features/CreateAcceptance.feature
@@ -18,7 +18,7 @@
 Feature: CreateAcceptance
 
   Scenario: Create a single node with multiple labels
-    Given any graph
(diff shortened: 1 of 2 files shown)


tck/features/CreateAcceptance.feature
```diff
@@ -18,7 +18,7 @@
 Feature: CreateAcceptance
 
   Scenario: Create a single node with multiple labels
-    Given any graph
+    Given an empty graph
```
Mats-SX: Right, this is necessary for the `labels` metric, as we don't know if the labels were in the graph or not.

PR #282, opened 2018-01-05
Fix side-effect handling and allow custom errors in tests

pstutz:
Also make error message point to the problematic feature/scenario
 
Builds on top of 1c8fda6 and 6abd25e

### DIFF
--- a/tools/tck/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
+++ b/tools/tck/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
@@ -46,8 +46,8 @@ object CypherTCK {
     * Provides all the scenarios in the openCypher TCK.
     *
     * @note While each scenario is unique, several scenarios could have the same name.
-    * This happens when the scenario is generated from a Gherkin Scenario Outline, in which case
-    * all variants of the Scenario Outline get the same name.
+    *       This happens when the scenario is generated from a Gherkin Scenario Outline, in which case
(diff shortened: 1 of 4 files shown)


tools/tck/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
```diff
@@ -140,39 +144,50 @@ object CypherTCK {
...
-          List(ExpectError(errorType, time, detail).validate(), SideEffects().fillInZeros)
+          val expectedError = ExpectError(errorType, time, detail)
+          if (shouldValidate) {
+            expectedError.validate match {
+              case None => // No problem
+              case Some(errorMessage) =>
+                throw new InvalidFeatureFormatException(
+                  s"""Invalid error format in scenario "${pickle.getName}" from feature "$featureName":
+                    $errorMessage
+                    If this is a custom error, then disable this validation with tag "@allowCustomErrors"""")
```
Mats-SX: How are the custom errors validated?

Mats-SX:
Rebased to get green build.

tools/tck/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
pstutz: The flag disables error format validation.

pstutz:
In the Neo4j acceptance tests there are errors that are not in the set of supported TCK errors. So we needed a flag to disable validation for them.

PR #283, opened 2018-01-08
Fix column ordering

pstutz:
Also fix column ordering when printing results


Mats-SX:
Rebased with only the extra commits

PR #285, opened 2018-01-10
TCK: Make 'LIMIT TO TWO HITS' test more explicit

DarthMax:
Currently the test `ReturnAcceptanceTest.Limit to two hits` is ambiguous. 
It selects all nodes from a graph and then limits this set to two with out an explicit order. The expectation however tests for specific nodes to be return although every combination would be valid.

This PR splits the above test into two tests. One test returns a set of equal elements and limits this set. Another test is equal to the original one but introduces a specific order on the nodes.

### DIFF
--- a/tck/features/ReturnAcceptance.feature
+++ b/tck/features/ReturnAcceptance.feature
@@ -43,7 +43,7 @@ Feature: ReturnAcceptanceTest
       LIMIT 2
       """
     Then the result should be:
-      | i             |
+      | i |
       | 1 |
       | 1 |
     And no side effects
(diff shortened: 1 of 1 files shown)


tck/features/ReturnAcceptance.feature
```diff
@@ -35,6 +35,20 @@ Feature: ReturnAcceptanceTest
...
   Scenario: Limit to two hits
+    Given an empty graph
+    When executing query:
+      """
+      UNWIND [1, 1, 1, 1, 1] AS i
+      RETURN i
+      LIMIT 2
+      """
+    Then the result should be:
+      | i             |
```
Mats-SX: Weird formatting

DarthMax: Fixed!

PR #291, opened 2018-01-17
oC grammar: make PropertyOrLabelsExpression more specific. Fixes #288

### LINKED ISSUE #288: Property lookup and label constraint expression
According to the grammar, `PropertyOrLabelsExpression` is built as

> `Atom` ( `PropertyLookup` | `NodeLabels` )*

https://github.com/opencypher/openCypher/blob/c1a86ad471c475fbc934a2564ee69a41a7889c0e/grammar/basic-grammar.xml#L258-L267

A few legal forms of this expression I know of are as follows:

 - filter by labels of the node variable
```
MATCH (p)
WHERE p:Person:Teacher
RETURN p
```

 - extract a node from a map, and optionally filter on its labels
```
MATCH (p)
WHERE {foo: p}.foo:Person:Teacher
RETURN p
```

But I know of no legal uses where `PropertyLookup` might follow `NodeLabels`, i.e. the following is illegal:
```
MATCH (p)
WHERE p:Person:Teacher.name
RETURN p
```

The error returned in Neo4j 3.2.3:

> Neo.ClientError.Statement.SyntaxError
> 
> Type mismatch: expected Any, Map, Node or Relationship but was Boolean (line 2, column 8 (offset: 17))
> "WHERE p:Person:Teacher.name"

This makes me to think of the following grammar form as a better solution, as `NodeLabels` is already a sequence of labels:

`Atom` `PropertyLookup`* `NodeLabels`?

https://github.com/opencypher/openCypher/blob/c1a86ad471c475fbc934a2564ee69a41a7889c0e/grammar/basic-grammar.xml#L149-L152

Is there any working examples or future plans that needs the current grammar for `PropertyOrLabelsExpression`, i.e. where a node label constraint is followed by a property lookup? If not, I would be happy to submit a PR to rewrite this part of the grammar.

jmarton:
The grammar used to allow mixing of property lookups and node labels
in a sequence. In reality, however, all property lookups perede all
the label constraints.

This grammar modification reflects the real structure.

### DIFF
--- a/grammar/basic-grammar.xml
+++ b/grammar/basic-grammar.xml
@@ -258,12 +258,9 @@
   <production name="PropertyOrLabelsExpression" rr:inline="true">
     <non-terminal ref="Atom"/>
     <repeat>
-      &WS;
-      <alt>
-        <non-terminal ref="PropertyLookup"/>
-        <non-terminal ref="NodeLabels"/>
-      </alt>
+      &WS; <non-terminal ref="PropertyLookup"/>
     </repeat>
+    <opt>&WS; <non-terminal ref="NodeLabels"/></opt>
   </production>
 
   <production name="Atom">
--- a/tools/grammar/src/test/java/org/opencypher/tools/grammar/Antlr4TestUtils.java
+++ b/tools/grammar/src/test/java/org/opencypher/tools/grammar/Antlr4TestUtils.java
@@ -133,9 +133,9 @@ public class Antlr4TestUtils
         }
 
         @Override
-        public void syntaxError( Recognizer<?,?> recognizer, Object o, int i, int i1, String s, RecognitionException e )
+        public void syntaxError( Recognizer<?,?> recognizer, Object o, int line, int charPositionInLine, String msg, RecognitionException e )
         {
-            fail( "syntax error in query: " + query );
+            fail( "syntax error in query at line " + line + ":" + charPositionInLine + ": " + msg + ": " + query );
         }
 
         @Override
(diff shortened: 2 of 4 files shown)


Mats-SX:
@jmarton It would be lovely if you could add queries in the test files https://github.com/opencypher/openCypher/blob/master/tools/grammar/src/test/resources/cypher-error.txt and https://github.com/opencypher/openCypher/blob/master/tools/grammar/src/test/resources/cypher.txt to verify that the grammar change works as intended. The test that uses those files is https://github.com/opencypher/openCypher/blob/master/tools/grammar/src/test/java/org/opencypher/tools/grammar/Antlr4ParserTest.java.

jmarton:
Thank you @Mats-SX for reminding to add test queries.

Working on the test queries also revealed as demonstrated in afbbd63a982fddb61207cdb4268f51cfbc329d43 that the original grammar of `PropertyOrLabelsExpression` was ambiguous as `NodeLabels`, i.e. a sequence of `NodeLabel` non-terminals was repeatable in the rule. It was ambiguous when parsing the query below,
 - if this is a sequence of two single-element NodeLabels match, i.e. `(:Person@NodeLabels, :Teacher@NodeLabels), or
 - a single match on a two-element NodeLabels, i.e. `(:Person, :Teacher)@NodeLabels`
    
```
MATCH (p)
WHERE p:Person:Teacher
RETURN p
```

My original rework of the grammar in d7ec4b8865740c9acb80c74818b3ebe3825db854 fixed this and the build again turned green when the demo and the fix were merged together.

PR #293, opened 2018-01-25
Close filesystem after parsing

pstutz:


pstutz:
I added a test. I was a bit tricky because the classpath access only works with Maven and not from within IntelliJ.

PR #304, opened 2018-03-01
Grammar simplifications

Mats-SX:
This change allows any updating clause to begin a query (part).

A continuation from #300 

A multi-part query now follows this simple outline:

1. one or more of
  a. zero or more reading clauses
  b. zero or more updating clauses
  c. a WITH clause
2. one SinglePartQuery

screen shot 2018-03-01 at 11 57 28

### DIFF
--- a/grammar/cypher.xml
+++ b/grammar/cypher.xml
@@ -106,7 +106,7 @@
       </seq>
       <seq>
         <non-terminal ref="UpdateOnlyQueryPartBody"/>
-        <opt>&WS; <non-terminal ref="Return"/></opt>
+        <opt><non-terminal ref="Return"/></opt>
       </seq>
     </alt>
   </production>
@@ -131,9 +131,8 @@
   </production>
 
   <production name="UpdateOnlyQueryPartBody" rr:inline="true">
-    <non-terminal ref="UpdatingStartClause"/>
-    <repeat>
-      &WS; <non-terminal ref="UpdatingClause"/>
+    <repeat min="1">
+      <non-terminal ref="UpdatingClause"/> &WS;
     </repeat>
   </production>
 
@@ -156,7 +155,6 @@
       </seq>
       <seq>
         <non-terminal ref="UpdateOnlyQueryPartBody"/>
-        &WS;
       </seq>
     </alt>
     <non-terminal ref="With"/>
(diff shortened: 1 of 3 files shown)


Mats-SX:
@jmarton What do you think of these changes? I intend to perform a new milestone release soon, and it would be nice to include these simplifications if you think they make sense.

jmarton:
@Mats-SX I think these simplifications indeed make sense from the grammar point of view, and they should be included in the next release, if we agree to go forward with them.

However, I think there is more to this simplification and I feel like we will end up logically reverting 3702d2eeb17380a2291c667c7c92072eb96253cc.

Unfortunately I can probably get back to this in more detail only on Monday. If it's more urgent, I can probably get back to this tomorrow.

Let me illustrate what I was thinking about.

Dropping the notion of `UodateStartClause` seems like eliminating the effective difference between `ReadUpdateQueryPartBody` and `UpdateOnlyQueryPartBody`, thus they seem to worth merging together e.g. in `ReadUpdateQueryPartBody`

Then there is no more substantial difference between a SingleQuery of the following forms, so they should be merged:
 - `ReadOnlyQueryPartBody RETURN`
 - `ReadUpdateQueryPartBody RETURN`

After this step, a single-part query would consist of
 a. zero or more reading clauses
 b. zero or more updating clauses
 c. a RETURN clause

As per your initial post, a multi-part query now follows this simple outline:

1. one or more of
    a. zero or more reading clauses
    b. zero or more updating clauses
    c. a WITH clause
2. one SinglePartQuery

So we have arrived to the following structure:

A query part body is
 a. zero or more reading clauses
 b. zero or more updating clauses

A single part query is
 a. one query part body
 b. a RETURN clause

A multi-part query is
 1. one or more of
    a. one query part body
    b. a WITH clause
 2. a single part query, which, in turn is
    a. one query part body
    b. a RETURN clause

Mats-SX:
@jmarton Thanks for the quick feedback. I think you are correct in identifying that there are additional simplifications that could be done, but I don't see any semantics being blocked by the current state on this PR (which does add over what is currently on `master`).

With regards to what a SinglePartQuery is, your outline is not quite correct. A SinglePartQuery still has two legitimate forms:

1. Zero or more reading clauses followed by one RETURN clause
2. Zero or more reading clauses, one or more updating clauses, zero or one RETURN clause

That is, a SingePartQuery doesn't _need_ to end in RETURN, in case the last clause is an updating clause.

I will make an attempt at simplifying the grammar structure following your observations.

jmarton:
Seems fine, @Mats-SX! Along with your commit of yesterday, I don't see anything missing, so this simplification worth merging.

You are right that I forgot that `RETURN` is optional for updating single queries.

PR #305, opened 2018-03-07
Fix faulty use of system default file separator

SaschaPeukert:
The previous version led to problems when running on windows where file.separator equals "\\"
( getClass.getResource does not handle "\\" )


tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
```diff
@@ -36,7 +36,7 @@ import scala.util.{Failure, Success, Try}
 
 object CypherTCK {
 
-  val featuresPath = s"${File.separator}features"
+  val featuresPath = "/features"
```
Mats-SX: Does this work on Windows? Isn't the file separator on Windows `\` (backslash)?

SaschaPeukert:
Please don't completly forget about this one :-)

PR #306, opened 2018-03-13
Addition of Attribution Notice to various files

petraselmer:


Mats-SX:
I think we need to modify this file: https://github.com/opencypher/openCypher/blob/master/ASL-2-header.txt and then run `mvn license:format` which would add the attribution notice to all files.

PR #310, opened 2018-04-05
List Operations TCK

petraselmer:
@Mats-SX please review and merge when you can. 

The new file `ListOperations` fully subsumes the following two files in the neo4j repo: 

- `NullListAcceptanceTest` 
- `ListExpressionAcceptanceTest`

I will make a note to the Cypher Team that they delete these two files once this PR has been merged and a new milestone created.

### DIFF
--- a/tck/features/ListOperations.feature
+++ b/tck/features/ListOperations.feature
@@ -123,7 +123,7 @@ Feature: ListOperations
       RETURN [1] IN [1, 2, [1]] AS res
       """
     Then the result should be:
-      | res   |
+      | res  |
       | true |
     And no side effects
 
@@ -133,7 +133,7 @@ Feature: ListOperations
       RETURN [1, 2] IN [1, [1, 2]] AS res
       """
     Then the result should be:
-      | res   |
+      | res  |
       | true |
     And no side effects
 
@@ -173,7 +173,7 @@ Feature: ListOperations
       RETURN [[1, 2], [3, 4]] IN [5, [[1, 2], [3, 4]]] AS res
       """
     Then the result should be:
-      | res   |
+      | res  |
       | true |
     And no side effects
 
@@ -183,7 +183,7 @@ Feature: ListOperations
       RETURN [[1, 2], 3] IN [1, [[1, 2], 3]] AS res
       """
     Then the result should be:
-      | res   |
+      | res  |
       | true |
     And no side effects
 
@@ -193,7 +193,7 @@ Feature: ListOperations
       RETURN [[1]] IN [2, [[1]]] AS res
       """
     Then the result should be:
-      | res   |
+      | res  |
       | true |
     And no side effects
 
@@ -203,7 +203,7 @@ Feature: ListOperations
       RETURN [[1, 3]] IN [2, [[1, 3]]] AS res
       """
     Then the result should be:
-      | res   |
+      | res  |
       | true |
     And no side effects
 
@@ -229,24 +229,24 @@ Feature: ListOperations
 
   # IN operator - null
 
-  Scenario: IN should return null if LHS and RHS null
+  Scenario: IN should return null if LHS and RHS are null
     When executing query:
       """
       RETURN null IN [null] AS res
       """
     Then the result should be:
-      | res   |
-      | null  |
+      | res  |
+      | null |
     And no side effects
 
-  Scenario: IN should return null if LHS and RHS null - list version
+  Scenario: IN should return null if LHS and RHS are null - list version
     When executing query:
       """
       RETURN [null] IN [[null]] AS res
       """
     Then the result should be:
-      | res   |
-      | null  |
+      | res  |
+      | null |
     And no side effects
 
   Scenario: IN should return null when LHS and RHS both ultimately contain null, even if LHS and RHS are of different types (nested list and flat list)
@@ -255,8 +255,8 @@ Feature: ListOperations
       RETURN [null] IN [null] AS res
       """
     Then the result should be:
-      | res   |
-      | null  |
+      | res  |
+      | null |
     And no side effects
 
   Scenario: IN with different length lists should return false despite nulls
@@ -275,7 +275,7 @@ Feature: ListOperations
       RETURN 3 IN [1, null, 3] AS res
       """
     Then the result should be:
-      | res   |
+      | res  |
       | true |
     And no side effects
 
@@ -285,7 +285,7 @@ Feature: ListOperations
       RETURN 4 IN [1, null, 3] AS res
       """
     Then the result should be:
-      | res   |
+      | res  |
       | null |
     And no side effects
 
@@ -295,7 +295,7 @@ Feature: ListOperations
       RETURN [1, 2] IN [[null, 'foo'], [1, 2]] AS res
       """
     Then the result should be:
-      | res   |
+      | res  |
       | true |
     And no side effects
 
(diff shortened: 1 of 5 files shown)


tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+    And no side effects
+
+  Scenario: Setting and returning the size of a list property
+    And having executed:
+      """
+      CREATE (:Label)
+      """
+    When executing query:
+      """
+      MATCH (n:Label)
```
Mats-SX: This scenario needs a different `Given` step. It may give more rows if given a graph with `:Label`-labeled nodes in it, which is allowed by `any graph`.

Each scenario may only have one `Given`, and the `Background` is overridden by what the scenario itself says, so we could just say `Given an empty graph` here.

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+    And no side effects
+
+  Scenario: Returning nested expressions based on list property
+    And having executed:
+      """
+      CREATE (:Label)
+      """
+    When executing query:
+      """
+      MATCH (n:Label)
```
Mats-SX: Because of the `MATCH` this can not be specified for `any graph`

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+    When executing query:
+      """
+      RETURN [1, 10, 100] + [4, 5] AS foo
+      """
+    Then the result should be:
+      | foo                |
+      | [1, 10, 100, 4, 5] |
+    And no side effects
+
+  Scenario: Appending lists of same type
```
Mats-SX: This scenario calls this `appending` and the above calls it `concatenating`. We should try to be consistent. I vote for `concatenating` because it isn't clear whether the operation is `appending` or `prepending`, which are both "directed".

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+    When executing query:
+      """
+      RETURN [false, true] + false AS foo
+      """
+    Then the result should be:
+      | foo                  |
+      | [false, true, false] |
+    And no side effects
+
+  Scenario: Execute n[0]
```
Mats-SX: We should call this `Indexing into literal list` and move it next to `Indexing into nested literal lists`

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+    And no side effects
+
+  Scenario: Extract eagerly
+    And having executed:
+      """
+      CREATE (:Label1 {name: 'original'})
+      """
+    When executing query:
+      """
+      MATCH (a:Label1)
```
Mats-SX: Needs `Given an empty graph` because of the `MATCH`

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+      WITH collect(a) AS nodes
+      WITH nodes, extract(x IN nodes | x.name) AS oldNames
```
Mats-SX: `extract()` is legacy and should not be part of the TCK. We can use a list comprehension here:

```
WITH nodes, [x IN nodes | x.name] AS oldNames
```

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+    When executing query:
+      """
+      RETURN [1, 2, 3][0] AS value
+      """
+    Then the result should be:
+      | value |
+      | 1     |
+    And no side effects
+
+  Scenario: Extract eagerly
```
Mats-SX: Rename to something like `collect and extract using a list comprehension`

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+      CREATE (:Label1 {name: 'original'})
+      """
+    When executing query:
+      """
+      MATCH (a:Label1)
+      WITH collect(a) AS nodes
+      WITH nodes, extract(x IN nodes | x.name) AS oldNames
+      UNWIND nodes AS n
+      SET n.name = 'newName'
+      RETURN n.name, oldNames
```
Mats-SX: This query does a whole lot of things -- doesn't seem very focussed in terms of what it tries to achieve

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+  Scenario: Filter eagerly
+    And having executed:
+      """
+      CREATE (:Label1 {name: 'original'})
+      """
+    When executing query:
+      """
+      MATCH (a:Label1)
+      WITH collect(a) AS nodes
+      WITH nodes, filter(x IN nodes WHERE x.name = 'original') AS noopFiltered
```
Mats-SX: `filter()` is legacy, we should use a list comprehension:

```
WITH nodes, [x IN nodes WHERE x.name = 'original'] AS noopFiltered
```

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+      RETURN n.name, oldNames
+      """
+    Then the result should be:
+      | n.name    | oldNames   |
+      | 'newName' | ['original'] |
+    And the side effects should be:
+      | +properties | 1 |
+      | -properties | 1 |
+
+  Scenario: Filter eagerly
```
Mats-SX: Not sure why it is named `eagerly`? I would prefer a more descriptive name

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+    And the side effects should be:
+      | +properties | 1 |
+      | -properties | 1 |
+
+  Scenario: Length on filter
+    When executing query:
+      """
+      MATCH (n)
+      OPTIONAL MATCH (n)-[r]->(m)
+      RETURN length(filter(x IN collect(r) WHERE x <> null)) AS cn
```
Mats-SX: `length()` should not be used for lists; we should use `size()`.
`filter()` should be replaced with a list comprehension.

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+      | n.name    | length(noopFiltered)   |
+      | 'newName' | 1                      |
+    And the side effects should be:
+      | +properties | 1 |
+      | -properties | 1 |
+
+  Scenario: Length on filter
+    When executing query:
+      """
+      MATCH (n)
```
Mats-SX: Needs `Given an empty graph`

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+      RETURN n.name, length(noopFiltered)
+      """
+    Then the result should be:
+      | n.name    | length(noopFiltered)   |
+      | 'newName' | 1                      |
+    And the side effects should be:
+      | +properties | 1 |
+      | -properties | 1 |
+
+  Scenario: Length on filter
```
Mats-SX: Should change name due to below comment

tck/features/ListOperations.feature
```diff
@@ -0,0 +1,748 @@
...
+      """
+      WITH $expr AS expr, $idx AS idx
+      RETURN expr[toInteger(idx)] AS value
+      """
+    Then the result should be:
+      | value |
+      | 'Apa' |
+    And no side effects
+
+  # Failures at runtime
```
Mats-SX: Perhaps we could partition the negative tests into its own feature? Perhaps `ListOperationsNegative.feature` or `ListOperationsErrors.feature`?

sherfert: I think they belong together and 3 scenarios also don't justify their own feature file. But I'll leave the last call to you @Mats-SX ;)

Mats-SX:
This looks ready for merge IMO. There is a test failure that requires investigation.

PR #317, opened 2018-06-15
Allow escaping labels and property keys for TCK Cypher Value Parser

DarthMax:


tools/tck-api/src/test/scala/org/opencypher/tools/tck/CypherValueVisitorTest.scala
```diff
@@ -46,13 +46,27 @@ class CypherValueVisitorTest extends FunSuite with Matchers {
...
+  }
+
   test("relationship") {
     val string = "[:A {since: 1920}]"
     val parsed = CypherValue(string)
     val expected = CypherRelationship("A", CypherPropertyMap(Map("since" -> CypherInteger(1920))))
     parsed should equal(expected)
   }
 
+  test("relationship with escaped label") {
```
Mats-SX: escaped `type`

tools/tck-api/src/main/resources/FeatureResults.g4
```diff
@@ -80,17 +80,17 @@ propertyMap : '{' (mapContents)? '}' ;
...
 
 propertyValue : value ;
 
-relationshipType : ':' relationshipTypeName ;
+relationshipType : ':' (relationshipTypeName | ('`' relationshipTypeName '`'));
 
 relationshipTypeName : SYMBOLIC_NAME ;
 
-label : ':' labelName ;
+label : ':' (labelName | ('`' labelName '`')) ;
```
Mats-SX: Since all these three rules resolve to the same lexer rule `SYMBOLIC_NAME`, perhaps we could simplify the change by introducing an in-between parser rule `symbolicName` that allows escaping?

Like

```
propertyKey: symbolicName ;
relationshipTypeName: symbolicName ;
labelName: symbolicName ;
symbolicName: '`' SYMBOLIC_NAME '`' ;
```

Mats-SX:
@DarthMax Is this still something we need?

Issue #319, opened 2018-07-17
Wrong validation query


swilly22:
Under SemanticErrorAcceptance.feature
Scenario: Failing when using MERGE on a node that is already bound

MATCH is used instead of MERGE.

PR #320, opened 2018-07-17
Accidentally used MATCH instead of MERGE.

swilly22:
See #319

### DIFF
--- a/tck/features/SemanticErrorAcceptance.feature
+++ b/tck/features/SemanticErrorAcceptance.feature
@@ -204,7 +204,7 @@ Feature: SemanticErrorAcceptance
   Scenario: Failing when using MERGE on a node that is already bound
     When executing query:
       """
-      MATCH (a)
+      MERGE (a)
       CREATE (a)
       """
     Then a SyntaxError should be raised at compile time: VariableAlreadyBound


tck/features/SemanticErrorAcceptance.feature
```diff
@@ -204,7 +204,7 @@ Feature: SemanticErrorAcceptance
   Scenario: Failing when using MERGE on a node that is already bound
     When executing query:
       """
-      MATCH (a)
+      MERGE (a)
```
petraselmer: Thanks for raising @swilly22 - this test is indeed erroneous, so well-spotted!
However, the test ought to be `MATCH (a) MERGE (a)`, as the test is checking that a failure occurs when attempting to merge a node that has already been found - via the match - and has thus been bound, meaning it is line 208 which is incorrect.

swilly22: Oh, sorry about that, I'll fix my PR.

petraselmer:
@swilly22 thanks very much for raising this PR. 

In order accept code contributions, the openCypher project requires submitters to agree to the openCypher Contributors Agreement (oCCA), details of which can be found [here](http://www.opencypher.org/contributions). It's a very quick process. Please ping me any questions.

tck/features/SemanticErrorAcceptance.feature
petraselmer: Super 👍

PR #323, opened 2018-08-14
Moved upload-release script to top level

sherfert:
The script would not work from the sub-directory.
Until this is fixed, it should live in the top-level.
This also fixes some small issues with the script.


Mats-SX:
I guess I was invoking it as `.build/upload-release.sh`. It's a bit ugly to not live in a directory, but I agree with the message.

PR #324, opened 2018-08-16
Split StringListNullOperationExpression to specific parts.

### LINKED ISSUE #321: Questions about StringListNullOperatorExpression
I was looking into the `StringListNullOperatorExpression` grammar clause with @marci543 and we had a few minor questions.

The relevant snippet of the grammar XML is the following:

https://github.com/opencypher/openCypher/blob/15157c0fe4243da1cbfe2c372bcd00fec77a5c17/grammar/basic-grammar.xml#L239-L263

1. The part `<seq> &WS; [ <opt>&expr;</opt> .. <opt>&expr;</opt> ] </seq>` would allow a list to be indexed as `some_list[..]`. This is not a valid expression. A workaround could be to enumerate all valid cases (only lower bound, only upper bound, both lower and upper bounds). This might be more trouble than worth though.

2. We didn't find the name `StringListNullOperatorExpression` very intuitive. Does it mean something like "string operator, list indexer or null check"?

jmarton:
This splits the StringListNullOperationExpression to subrules in the grammar to separate string, list and null operations.

I think this way of separation to subrules has no consequences on the operator precedences. @Mats-SX, @szarnyasg please take a look at it.

Should close #321

### DIFF
--- a/grammar/basic-grammar.xml
+++ b/grammar/basic-grammar.xml
@@ -244,22 +244,38 @@
   <production name="StringListNullOperatorExpression" rr:inline="true">
     <non-terminal ref="PropertyOrLabelsExpression"/>
     <repeat><alt>
+      <non-terminal ref="StringOperatorExpression"/>
+      <non-terminal ref="ListOperatorExpression"/>
+      <non-terminal ref="NullOperatorExpression"/>
+    </alt></repeat>
+  </production>
+
+  <production name="ListOperatorExpression" rr:inline="true">
+    <alt>
       <seq> &WS; [ &expr; ] </seq>
       <seq> &WS; [ <opt>&expr;</opt> .. <opt>&expr;</opt> ] </seq>
-      <seq>
-        <alt>
-          <non-terminal ref="RegularExpression"/>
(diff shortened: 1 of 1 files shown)


szarnyasg:
The relevant part of the generated ANTLR code looks like this:

```
oC_StringListNullOperatorExpression : oC_PropertyOrLabelsExpression ( oC_StringOperatorExpression | oC_ListOperatorExpression | oC_NullOperatorExpression )* ;

oC_StringOperatorExpression : ( ( SP IN ) | ( SP STARTS SP WITH ) | ( SP ENDS SP WITH ) | ( SP CONTAINS ) ) SP? oC_PropertyOrLabelsExpression ;

oC_ListOperatorExpression : ( SP? '[' oC_Expression ']' )
                          | ( SP? '[' oC_Expression? '..' oC_Expression? ']' )
                          ;

oC_NullOperatorExpression : ( SP IS SP NULL )
                          | ( SP IS SP NOT SP NULL )
                          ;
```

(I removed `IN : ( 'I' | 'i' ) ( 'N' | 'n' )`-style rules and changed the order.)

This looks good to me.

PR #328, opened 2018-08-30
Type filtering for IBO

hvub:
1st Draft of a CIP for type filter functions.

### DIFF
--- a/cip/1.accepted/CIP2018-08-27-type-filtering-for-IBO.adoc
+++ b/cip/1.accepted/CIP2018-08-27-type-filtering-for-IBO.adoc
@@ -73,12 +73,12 @@ Hence, this CIP proposes to extended Cypher with dedicated type-filtering predic
 Following the link:https://neo4j.com/docs/developer-manual/3.4/cypher/syntax/values/#property-types[property type categorization] of Cypher, we define a _property type category_ as being one of the categories listed in the following table.
 
 |===
-| *Category* | *Types*
-| Number     | Integer, Float
-| String     | String
-| Boolean    | Boolean
-| Spatial    | Point
-| Temporal   | Date, Time, LocalTime, DateTime, LocalDateTime, Duration
+| *Category*      | *Types*
+| Number          | Integer, Float
+| String          | String
+| Boolean         | Boolean
+| Spatial         | Point
+| Temporal        | Date, Time, LocalTime, DateTime, LocalDateTime, Duration
 | ListOfNumbers   | List[Number]
 | ListOfStrings   | List[String]
 | ListOfBooleans  | List[Boolean]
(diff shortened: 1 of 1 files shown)


cip/1.accepted/CIP2018-08-27-type-filtering-for-IBO.adoc
```diff
@@ -0,0 +1,166 @@
...
+| String          | `isString(<exp>)`
+| Boolean         | `isBoolean(<exp>)`
+| Spatial         | `isSpatial(<exp>)`
+| Temporal        | `isTemporal(<exp>)`
+| ListOfNumbers   | `isListOfNumbers(<exp>)`
+| ListOfStrings   | `isListOfStrings(<exp>)`
+| ListOfBooleans  | `isListOfBooleans(<exp>)`
+| ListOfSpatials  | `isListOfSpatials(<exp>)`
+| ListOfTemporals | `isListOfTemporals(<exp>)`
+|===
```
technige: Alternative naming for (e.g.) `isString(x) could be `string.is(x)`. This could be complemented by renaming `toString(x)` to `string.cast(x)`. List functions could perhaps use `string.isList(x)` or even `string.list.is(x)`.

This naming would be consistent with some of the new `date.foo` and `time.foo` functions added alongside the temporal types. It would also allow functions relating to the same type to appear side-by-side in a drop-down list.

Mats-SX: I kinda like this way of namespacing functions. One comment though is that I'd suggest `list.string.is(x)` for the function that checks if `x` is a `LIST OF STRING`. It's about a list so it makes sense to have that word first.

alastai: Temporal functions don't use the suggested pattern for type casts. They are of the general form `<type>.<qualifier>(<arguments>)` where a cast is `<type>(<expression>)`: e.g. `datetime('2015-07-21T21:40:32.142+0100')`. To follow the same pattern would indicate that the function for casting to a string would be `string(x)`.

alastai: SQLite has the same "dynamic typing" (instance vs schema typing) as Cypher. They resolve the sort order issue differently. See https://www.sqlite.org/datatype3.html.

I am interpreting the text of this CIP to mean: "if the result of a function for type X is false then ignore the row concerned in order to ensure that the rows returned have a property which is uniformly of type X or NULL".

If that is a correct understanding, then I believe we should find a way of preserving all the rows in the result set, such that the property concerned is are known to be of type X, or is of another type Y (to be sorted appropriately), or are NULL. This approach would mirror the existing fixed sort order of NULL with respect to null/absent-valued properties, would not require a syntactic change, and would prevent cardinality and value changes when comparing unsorted and sorted result sets.

alastai: Correcting two mistakes in my comment of yesterday. 

1) SQLite and the Cypher [global sort order](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc) share the same type ordering approach. 

2) But this is irrelevant to the purpose of the proposal, as it aims to impose homogeneous typing by subsetting only those properties of a given type: i.e. the proposal filters out only true results, not true plus NULL.

thobe:
Neo4j are withdrawing this proposal for now. Our current thinking is that we will pursue the notion of types from a different angle.

PR #329, opened 2018-09-21
TCK scenarios for list slices

dwitry:
- Acceptance tests for list slice
- Covers valid and invalid ranges
- Ported from Cypher for Gremlin

### DIFF
--- a/tck/features/ListOperations.feature
+++ b/tck/features/ListOperations.feature
@@ -873,19 +873,27 @@ Feature: ListOperations
       | [1, 2, 3] |
     And no side effects
 
-  Scenario: List slice with null range
+  Scenario Outline: List slice with null range
     Given any graph
     When executing query:
-      """
-      WITH [1, 2, 3] AS list
-      RETURN list[1..null] AS r
-      """
+        """
+        WITH [1, 2, 3] AS list
+        RETURN list[<lower>..<upper>] AS r
+        """
     Then the result should be:
       | r    |
       | null |
     And no side effects
 
-  Scenario: List slice with parametrized range
+    Examples:
+      | lower | upper |
+      | null  | null  |
+      | 1     | null  |
+      | null  | 3     |
+      |       | null  |
+      | null  |       |
+
+  Scenario: List slice with parameterised range
     Given any graph
     And parameters are:
       | from | 1 |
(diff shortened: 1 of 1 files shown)


Mats-SX:
[Review PR #329 TCK scenarios for list slices](https://trello.com/c/PNM4LFxu/82-review-pr-329-tck-scenarios-for-list-slices)

tck/features/ListOperations.feature
```diff
@@ -775,6 +775,146 @@ Feature: ListOperations
...
+      """
+      WITH [1, 2, 3] AS list
+      RETURN list[1..null] AS r
+      """
+    Then the result should be:
+      | r    |
+      | null |
+    And no side effects
+
+  Scenario: List slice with parametrized range
```
Mats-SX: parametrized -> parameterised

tck/features/ListOperations.feature
```diff
@@ -775,6 +775,146 @@ Feature: ListOperations
...
+      """
+      WITH [1, 2, 3] AS list
+      RETURN list[$from..$to] AS r
+      """
+    Then the result should be:
+      | r      |
+      | [2, 3] |
+    And no side effects
+
+  Scenario: List slice with parametrized invalid range
```
Mats-SX: parametrized -> parameterised

tck/features/ListOperations.feature
```diff
@@ -775,6 +775,146 @@ Feature: ListOperations
...
+    Given any graph
+    When executing query:
+      """
+      WITH [1, 2, 3] AS list
+      RETURN list[1..null] AS r
+      """
+    Then the result should be:
+      | r    |
+      | null |
+    And no side effects
```
Mats-SX: Could the `null` case be expanded to a `Scenario Outline` here, where we try all combinations of `null` arguments to the list slice?

Something like

```
  Scenario Outline: List slice with null range
    Given any graph
    When executing query:
      """
      WITH [1, 2, 3] AS list
      RETURN list[<lower>..<upper>] AS r
      """
    Then the result should be:
      | r    |
      | null |
    And no side effects

    Examples:
      | lower | upper | 
      | null    | null    | 
      | 1    | null    | 
      | null    | 3    | 
      | | null | // empty string -- does this even work? 
      | null | | 
```

or with a few standard scenarios if the empty string thing doesn't work or you don't like the outline. The point is that there are a few more cases with `null` that could be covered.

PR #332, opened 2018-10-19
Specify the behaviour of deleted elements

thobe:
It has become evident that there are some inconsistencies and unhelpful behaviour around the handling of deleted elements. This proposal aims to rectify that.

### DIFF
--- a/cip/1.accepted/CIP2018-10-19-Delete-Semantics.adoc
+++ b/cip/1.accepted/CIP2018-10-19-Delete-Semantics.adoc
@@ -1,0 +1,49 @@
+= CIP2018-10-19 Semantics of Deleted Elements
+:numbered:
+:toc:
+:toc-placement: macro
+:source-highlighter: codemirror
+
+*Author:* Tobias Lindaaker <tobias.lindaaker@neotechnology.com>
+
+toc::[]
+
+== Semantics of Deleted Elements
+
+Cypher allows read statements to occur after update statements.
+This includes reading statements after statements that delete elements.
+Since the driving table of the preceding query is retained into the succeeding reading statements, this means that entries in the driving table that previously contained elements might now contain elements that have been deleted.
+
+The semantics of accessing such elements is the same as accessing `NULL` values.
+
+This can be thought of as replacing all occurrences of the deleted elements (anywhere) in the driving table (including in nested values) with `NULL`, or as treating the deleted elements as _effectively_ `NULL`.
+
+These semantics effect the `DELETE` statement itself, even if not succeeded by further read statements, since the same element can occur in multiple rows in the driving table.
+
+
+=== Accessing properties of deleted Elements
+
+Accessing properties of deleted elements produces a `NULL` value, just like accessing a property from a `NULL` value would.
+
+=== Pattern matching on deleted Elements
+
+Pattern matching on deleted elements produces a single row containing only `NULL` bindings, in the same way as `OPTIONAL MATCH` would.
+
+=== Deleting deleted Elements
+
+Deleting a deleted element (or any `NULL` value) is a no-op.
+
+=== Equality of deleted Elements
+
+The normal semantics is that two `NULL` values are never considered equal.
+This extends to deleted elements, since they are equivalent to `NULL` for _all_ intents and purposes.
+
+[source, cypher]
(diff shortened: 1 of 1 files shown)


cip/1.accepted/CIP2018-10-19-Delete-Semantics.adoc
```diff
@@ -0,0 +1,49 @@
...
+Since the driving table of the preceding query is retained into the succeeding reading statements, this means that entries in the driving table that previously contained elements might now contain elements that have been deleted.
+
+The semantics of accessing such elements is the same as accessing `NULL` values.
+
+This can be thought of as replacing all occurrences of the deleted elements (anywhere) in the driving table (including in nested values) with `NULL`, or as treating the deleted elements as _effectively_ `NULL`.
+
+These semantics effect the `DELETE` statement itself, even if not succeeded by further read statements, since the same element can occur in multiple rows in the driving table.
+
+
+=== Accessing properties of deleted Elements
```
thobe: This is a change from current behaviour.

This is the main change that people seem to agree that we want from this change proposal.

cip/1.accepted/CIP2018-10-19-Delete-Semantics.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+Accessing properties of deleted elements produces a `NULL` value, just like accessing a property from a `NULL` value would.
+
+=== Pattern matching on deleted Elements
```
thobe: This is a change from current behaviour.

Currently pattern matching on `NULL` produces 0 rows.

cip/1.accepted/CIP2018-10-19-Delete-Semantics.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+Pattern matching on deleted elements produces a single row containing only `NULL` bindings, in the same way as `OPTIONAL MATCH` would.
+
+=== Deleting deleted Elements
```
thobe: This is the same as current behaviour, but should still be documented.

cip/1.accepted/CIP2018-10-19-Delete-Semantics.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+Deleting a deleted element (or any `NULL` value) is a no-op.
+
+=== Equality of deleted Elements
```
thobe: This is a change from current behaviour, where two references to the same element are considered equal even after the element has been deleted.

cip/1.accepted/CIP2018-10-19-Delete-Semantics.adoc
```diff
@@ -0,0 +1,69 @@
...
+.This query returns `same1: *true*; same2: *false*` for all rows
+----
+MATCH (n), (m)
+WHERE n = m AND NOT EXISTS { (n)-() }
+WITH n, m, n = m AS same1
+DELETE n
+RETURN same1, n = m AS same2
+----
+
+=== Deleted elements in paths
```
thobe: This is also a change from current behaviour.

The controversial part of it is that it could be argued that it might be expensive to implement such checking to see if the elements of the path have been deleted. I would argue that since all operations on a path accesses all the elements anyhow, the cost of such checking is negligible.

thobe: A nice thing about this behaviour for paths is that it works really well with the (intended) behaviour of `UNWIND`, where for example `UNWIND nodes(p) AS n` would produce a single row with `n` bound to `NULL` if an element of `p` was deleted (since `p` would be considered `NULL` and `nodes(NULL)` returns `NULL`, and `UNWIND NULL AS n` produces a single row with `n` bound to `NULL`).

cip/1.accepted/CIP2018-10-19-Delete-Semantics.adoc
sherfert: Why do we deviate from the "behaves like null" in this case?

thobe: The proposal in this section is to have this be the behaviour for _all_ nulls.

PR #334, opened 2018-10-29
Added CIP2018-10-29 EXISTS and IS NOT NULL

sherfert:
[CIP](https://github.com/sherfert/openCypher/blob/CIP-exists-is-not-null/cip/1.accepted/CIP2018-10-29-EXISTS-and-IS-NOT-NULL.adoc)

### DIFF
--- a/cip/1.accepted/CIP2018-10-29-EXISTS-and-IS-NOT-NULL.adoc
+++ b/cip/1.accepted/CIP2018-10-29-EXISTS-and-IS-NOT-NULL.adoc
@@ -11,6 +11,7 @@
 --
 This CIP limits the scope of `EXISTS` to existential subqueries and removes the syntax for property existence check.
 This check is enabled by `IS NOT NULL`.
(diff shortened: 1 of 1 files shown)


boggle:
Please always add a link to the description to the rendered version! That helps people to review CIPs

boggle:
Otherwise great CIP, I'm torn on the constraints change. Probably good to keep as is here but we need to look at those more holistically in a different CIP. A good next step might be to have a discussion about having a shrothand notation for `IS NOT NULL` and wether we want to add this here.

Mats-SX:
@sherfert I would like this CIP to also rewrite the section in the CIP it refers to changing. At the very least, insert a forward-reference pointing to this CIP. This is in order to not confuse readers of the older CIP.

thobe:
Accepted by lazy consensus: https://groups.google.com/forum/#!topic/opencypher/cAccRlkHdp8

PR #341, opened 2018-11-26
Mark extract and filter functions legacy in the grammar

### LINKED ISSUE #340: Mark `extract()` and `filter()` as legacy in grammar
As discussed in #228 `extract()` and `filter()` have been deprecated in favour of _list comprehensions_: `[var IN listExpr WHERE predicate]`.

The grammar should mark them as legacy properly.

szarnyasg:
@Mats-SX I took a stab at updating the grammar to fix #340. Please review.

### DIFF
--- a/grammar/basic-grammar.xml
+++ b/grammar/basic-grammar.xml
@@ -299,8 +299,7 @@
       <seq>COUNT &WS; ( &WS; * &WS; )</seq>
       <non-terminal ref="ListComprehension"/>
       <non-terminal ref="PatternComprehension"/>
-      <seq>FILTER  &WS; ( &WS; <non-terminal ref="FilterExpression"/> &WS; ) </seq>
-      <seq>EXTRACT &WS; ( &WS; <non-terminal ref="FilterExpression"/> &WS; <opt>&WS; | &expr;</opt> ) </seq>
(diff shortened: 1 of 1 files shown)


grammar/basic-grammar.xml
```diff
@@ -365,6 +365,14 @@
     <opt>&WS;<non-terminal ref="Where"/></opt>
   </production>
 
+  <production name="LegacyFilterExpression" rr:inline="true" oc:legacy="true">
+    <seq>FILTER  &WS; ( &WS; <non-terminal ref="FilterExpression"/> &WS; ) </seq>
+  </production>
+
+  <production name="LegacyExtractExpression" rr:inline="true" oc:legacy="true">
+    <seq>EXTRACT &WS; ( &WS; <non-terminal ref="FilterExpression"/> &WS; <opt>&WS; | &expr;</opt> ) </seq>
+  </production>
```
Mats-SX: How about we group these two under the same production? They are extremely similar (as we can see by them being replaced by the same, more general thing). We could call that production f.e. `LegacyListExpression` or `LegacyFilterExpression` and then we can move things there that become deprecated. WDYT?

szarnyasg: It makes sense, I changed the rule to `LegacyListExpression`.

PR #342, opened 2018-12-11
Model TCK index and categorisation

Mats-SX:
Attempts to:

- Have more feature files with a more categorised layout
- Have more structure among feature files
- Achieve higher visibility of the TCK's coverage of Cypher features
- Numbering scheme for scenarios


thobe:
I see the value in having numbers for each scenario in each feature file, that makes it easier to reference the scenarios. I am not convinced that this should extend to the naming of feature files, I think these would be better served having short descriptive names than numbers.

Taking the Create feature files as an example, it seems natural to change these to something like:

* `Create1` -> `Create.Nodes`
* `Create2` ->  `Create.Relationships`
* `Create3` -> `Create.Integration`
* `Create4` -> `Create.Failures`

petraselmer:
I think the file-naming issue is still open: i.e. once we have done a few of these, we will see the "shape" of the structure as a whole and be able to make a decision on this. My recollection was that at the beginning, we had considered descriptive names, but deemed that they would quickly become unwieldy and become a burden to use, and messy to read. 

However, that may indeed have been premature. I think we will need to revisit and review once we have covered enough ground to be able to have enough data points. 

It goes without saying that **good** descriptive file names trump numeric ones any day of the week...

Mats-SX:
I like the namespacing with `.`; nice style. I believe @hvub voiced a similar idea on going for descriptive names.

PR #349, opened 2019-02-12
allow WHERE directly after YIELD

### LINKED ISSUE #344: Parse error of "CALL-WHERE-RETURN" clause using ANTLR4 grammar
When I parse the query:
```
CALL db.labels() YIELD label
WHERE label CONTAINS 'User'
RETURN count(label) AS numLabels
```
I get: 
`line 9:0 no viable alternative at input '\nWHERE'`

It seems there is no rule to match the above query?

hvub:
allow YIELD *
forbid YIELD -
editorial change in InQueryCall

Co-Authored-By: Mats Rydberg <mats@neotechnology.com>

Fixes #344

### DIFF
--- a/grammar/cypher.xml
+++ b/grammar/cypher.xml
@@ -251,7 +251,11 @@
   </production>
 
   <production name="InQueryCall">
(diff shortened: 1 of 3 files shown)


Mats-SX:
Railroad diagram looks like this:

screen shot 2019-02-12 at 15 20 21

thobe:
Great! I wonder how `YIELD -` got past review, I remember that being a discussion point that we never actually intended to include in the grammar.

PR #360, opened 2019-03-29
Contribute Cypher for Gremlin TCK reporting

dwitry:
See README.adoc


pom.xml
```diff
@@ -32,7 +32,7 @@
     <dep.apiguardian.version>1.0.0</dep.apiguardian.version>
     <dep.batik.version>1.8</dep.batik.version>
     <dep.cucumber.version>4.2.0</dep.cucumber.version>
-    <dep.gherkin.version>5.0.0</dep.gherkin.version>
+    <dep.gherkin.version>5.1.0</dep.gherkin.version>
```
dwitry: Cucumber `4.2.0` has dependency on Gherkin `5.1.0` not `5.0.0`

tools/tck-reporting/src/main/java/org/opencypher/tools/tck/reporting/SystemOutReader.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.OutputStream;
+import java.io.PrintStream;
+
+class SystemOutReader {
+    private PrintStream systemOut = System.out;
+    private ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+    private TeeOutputStream teeOutput = new TeeOutputStream(System.out, buffer);
+
+    SystemOutReader() {
+        System.setOut(new PrintStream(teeOutput));
```
sherfert: Is there a nicer way to achieve the same other than changing the default output stream?

I wonder how this behaves when different test classes are run in parallel.

tools/tck-reporting/README.adoc
```diff
@@ -0,0 +1,28 @@
...
+    // See example in TCK Api
+  }
+----
+3. Run test with property `cucumber.options` = `--plugin PLUGIN[:PATH_OR_URL]`
+
+NOTE: `@cucumber.api.CucumberOptions` annotation is not recognized.
+
+== Use cases
+
+* Run with `-Dcucumber.options="--plugin json:/path/to/html"` to generate Cucumber html report
```
sherfert: I think the `json` in this line is misplaced.

dwitry:
> I still need to do a more thorough review. For now, I am getting an Exception when I try to run neo4j with this Adapter:

@sherfert, this error means that test events are coming it unexpected order. Please provide details of your setup, so I could try reproduce this.

tools/tck-reporting/src/main/java/org/opencypher/tools/tck/reporting/SystemOutReader.java
dwitry: 1. My understanding is that TCK is not designed to be executed in parallel.
2. Capturing the output stream is the most universal approach. Otherwise, it would require implementer to forward logs to the adapter, which is not very straight forward approach.
3. I was unaware of [new experimental output capturing functionality](https://junit.org/junit5/docs/current/user-guide/#running-tests-capturing-output) in JUnit 5.3, will look into it.

tools/tck-reporting/README.adoc
dwitry: Thank you for noticing. Fixed.

sherfert:
I had to make sure to not run the tests in forked mode, then the Exception disappears.

```
<plugin>
      <groupId>org.apache.maven.plugins</groupId>
      <artifactId>maven-surefire-plugin</artifactId>
      <configuration>
        <forCount>1</forkMode>
      </configuration>
    </plugin>
```

Can you please mention this in the README?

tools/tck-reporting/src/main/java/org/opencypher/tools/tck/reporting/SystemOutReader.java
dwitry: JUnit output capturing functionality, will not work in this case because the adapter is based on TCK events, not JUnit events.

I can't think of an approach that will be nicer than changing the default output stream. Maybe you have any ideas?

PR #362, opened 2019-04-05
Update reserved words list

Mats-SX:
- Adds `CALL`, `YIELD`, `MANDATORY` to Clauses list
- Adds `FROM`, `GRAPH` to list of future reserved words


Mats-SX:
@thobe I must've missed your review -- apologies. I agree with those remarks and I will just push a fix commit directly as this PR was merged recently (which brought my attention back to it).

Please see https://github.com/opencypher/openCypher/commit/4661834a63a19317cf580858157e4a5e7bb85587 and let me know if this was not acceptable.

thobe:
@Mats-SX looks good, thanks!

PR #363, opened 2019-04-09
Modernise integrity tests

Mats-SX:


hvub:
This work has be outdate by the #478.

#478 still uses the AntlrParser for the grammar check of queries, but consumes the scenarios through the TCK API which parser parameters and result sets through the Fastparse parser.

PR #364, opened 2019-04-11
Support escaping in string literals

Mats-SX:
Allows backticks to be used in string values


tools/tck-api/src/main/scala/org/opencypher/tools/tck/values/CypherValueParser.scala
```diff
@@ -102,7 +102,10 @@ class CypherValueParser(val orderedLists: Boolean) {
     }
 
   private def string[_: P]: P[CypherString] =
-    P("'" ~ CharsWhile(_ != ''', 0).!.map(CypherString) ~ "'")
+    P("'" ~/ (stringChunk | backslash | escape).rep.!.map { s =>
```
s1ck: Using your tests, I just tried to write my own parser to understand what's going on. I used some input from https://gitter.im/lihaoyi/fastparse/archives/2015/06/29 
I came to the same result :D So, I guess this is fine.

PR #366, opened 2019-05-15
build on windows - avoid backslash n

PeterFurniss:
Building on windows failed because some of the grammar tools and tests of same assumed System.lineSeparator was \n. Corrected to use lineSeparator or println as appropriate

Also the separation of example cypher queries in a test needs to specify that file character set is UTF-8 if the System default is different

OutputTest combined 3 assertions into one method call, but the Windows compiler failed to find that method. Split into three calls

### DIFF
--- a/tools/grammar/src/main/java/org/opencypher/tools/grammar/Antlr4.java
+++ b/tools/grammar/src/main/java/org/opencypher/tools/grammar/Antlr4.java
@@ -73,11 +73,11 @@ public class Antlr4 extends BnfWriter
         String header = grammar.header();
         if ( header != null )
         {
-            output.append( "/*\n * " )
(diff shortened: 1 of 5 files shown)


tools/grammar/src/main/java/org/opencypher/tools/grammar/Antlr4.java
```diff
@@ -128,14 +128,14 @@ public void close()
                 set.accept( new SetFormatter( output ) );
                 output.append( ']' );
             }
-            output.println( " ;\n" );
+            output.println( " ;" ).println();  // PRF was " ;\n"
```
thobe: Good change, unnecessary comment.

PR #369, opened 2019-06-19
Consistent property type refactoring

dwitry:
- Removing idle data type variations between properties of the same name across the scenarios
- Replace potential reserved keywords like `Label`, `key`

After this PR is reviewed, I will create one for categorization (#342).

# Summary

## Properties

* `id` → 
  - Integer (58) → 
  - Variable type (2): `id` → `var`
* `foo` →
  - Integer (3): `foo` → `num`
  - List<Integer> (4): `foo` → `numbers`
  - List<Map> (1): `foo` → `maplist`
  - String (6): `foo` → `name`
* `prop` →
  - Float (24): `prop` → `price`
  - String (14): `prop` → `name`
  - Integer (87): `prop` → `num`
  - Temporal (28): `prop` → `date`
  - Variable type (1): `prop` → `var`
* `x` →
  - Integer (11): `x` → `num`
  - Variable type (1): `x` → `var`
  - List<Integer> (2): `x` → `numbers`
* `y` →
  - Variable type (1): `y` → `var`
  - String (6): `y` → `name`
* `a` →
  - String (1): `a` → `name`
  - List<Integer> (2): `a` → `numbers`
* `prop1`,`prop2` →
  - Integer (3): `prop1` → `num`
  - List<Integer> (1): `prop1` → `list1`,  `prop2` → `list2`
* `age` →
  - Integer (9): →
  - String (1): `num` → `name`
* `p` →
  - Temporal (3): `p` → `date`
  - Temporal (3): `p` → `created`
  - List<Temporal> (2): `p` → `dates`
  - Integer (2): `p` → `num`
* `bar` →
  - Integer (3): `foo` → `num`
  - String (12): `bar` → `name2`
* `property` →
  - Integer: `property` → `num`
  - String: `property` → `name`
* `property1` →
  -  Integer: `property1` → `num1`
* `property2` → 
  - Integer: `property2` → `num2`
* `value` →
  - String (4): `value` → `name`
  - Integer (15): `value` → `num`
* `level`
  - `level` → `status`
* `created`
  - `created` → `ok`

## Reserved keywords

* `Label` → `TheLabel`
* `key` → `num`

### DIFF
--- a/tck/features/MatchAcceptance2.feature
+++ b/tck/features/MatchAcceptance2.feature
@@ -57,20 +57,20 @@ Feature: MatchAcceptance2
     And having executed:
       """
       CREATE (root:Root {name: 'x'}),
-             (child1:TextNode {id: 'text'}),
-             (child2:IntNode {id: 0})
+             (child1:TextNode {var: 'text'}),
+             (child2:IntNode {var: 0})
       CREATE (root)-[:T]->(child1),
              (root)-[:T]->(child2)
       """
     When executing query:
       """
       MATCH (:Root {name: 'x'})-->(i:TextNode)
-      WHERE i.id > 'te'
+      WHERE i.var > 'te'
       RETURN i
       """
     Then the result should be:
-      | i                        |
-      | (:TextNode {id: 'text'}) |
+      | i                         |
+      | (:TextNode {var: 'text'}) |
     And no side effects
 
   Scenario: Do not fail when evaluating predicates with illegal operations if the OR'd predicate evaluates to true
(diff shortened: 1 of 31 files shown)


tck/features/MatchAcceptance2.feature
```diff
@@ -578,31 +578,31 @@ Feature: MatchAcceptance2
...
       MATCH (n)
-      WHERE n.foo = 'bar'
+      WHERE n.name = 'bar'
       RETURN n
       """
     Then the result should be:
-      | n              |
-      | ({foo: 'bar'}) |
+      | n                |
+      | ({name: 'bar'}) |
```
thobe: looks like one space too many added in the first row

dwitry:
> It would be nice to have some sort of rules written down

There is [card for that](https://trello.com/c/oUZM6Xux/117-guidline-for-how-to-write-a-tck-scenario). If we agree on current naming, I will write it down.

>  or even automatically enforced for how things should be named.

I've created a script that shows the type for each property across all TCK tests. With some modifications, it should be possible to add validations to enforce single type for each property name (or `Any` type) and integrate into openCypher build pipeline. Currently it is an ad hoc solution with dependency on CfoG, but it could be improved.

> We might also want tests that assert that a system that does not have an explicit schema

Currently such scenarios have a property named `var`. The proper way to achieve this would be [tags](https://trello.com/c/RDUro9dL/101-tags-idea-from-dmitry).

tck/features/MatchAcceptance2.feature
dwitry: Thanks for noticing. Fixed.

PR #372, opened 2019-07-08
Correct wrong assertions for SKIP/LIMIT.

sherfert:
This corrects the wrong assertion that SKIP/LIMIT should succeed with
negative integers at runtime, to make it consistent with the compile
time behavior.

It adds more tests and asserts on the same behavior for floating point
numbers.

### DIFF
--- a/tck/features/SkipLimitAcceptance.feature
+++ b/tck/features/SkipLimitAcceptance.feature
@@ -30,8 +30,10 @@
 
 Feature: SkipLimitAcceptanceTest
 
-  Scenario: SKIP with an expression that depends on variables should fail
+  Background:
     Given any graph
+
+  Scenario: SKIP with an expression that depends on variables should fail
     When executing query:
       """
       MATCH (n) RETURN n SKIP n.count
@@ -39,7 +41,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at compile time: NonConstantExpression
 
   Scenario: LIMIT with an expression that depends on variables should fail
-    Given any graph
     When executing query:
       """
       MATCH (n) RETURN n LIMIT n.count
@@ -47,7 +48,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at compile time: NonConstantExpression
 
   Scenario: SKIP with an expression that does not depend on variables
-    Given any graph
     And having executed:
       """
       UNWIND range(1, 10) AS i
@@ -67,7 +67,6 @@ Feature: SkipLimitAcceptanceTest
 
 
   Scenario: LIMIT with an expression that does not depend on variables
-    Given an empty graph
     And having executed:
       """
       UNWIND range(1, 3) AS i
@@ -85,7 +84,6 @@ Feature: SkipLimitAcceptanceTest
     And no side effects
 
   Scenario: Negative parameter for LIMIT should fail
-    Given any graph
     And having executed:
       """
       CREATE (s:Person {name: 'Steven'}),
@@ -102,7 +100,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at runtime: NegativeIntegerArgument
 
   Scenario: Negative parameter for LIMIT with ORDER BY should fail
-    Given any graph
     And having executed:
       """
       CREATE (s:Person {name: 'Steven'}),
@@ -119,7 +116,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at runtime: NegativeIntegerArgument
 
   Scenario: Negative LIMIT should fail
-    Given any graph
     And having executed:
       """
       CREATE (s:Person {name: 'Steven'}),
@@ -134,7 +130,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at compile time: NegativeIntegerArgument
 
   Scenario: Negative parameter for SKIP should fail
-    Given any graph
     And having executed:
       """
       CREATE (s:Person {name: 'Steven'}),
@@ -151,7 +146,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at runtime: NegativeIntegerArgument
 
   Scenario: Negative SKIP should fail
-    Given any graph
     And having executed:
       """
       CREATE (s:Person {name: 'Steven'}),
@@ -166,7 +160,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at compile time: NegativeIntegerArgument
 
   Scenario: Floating point parameter for LIMIT should fail
-    Given any graph
     And having executed:
       """
       CREATE (s:Person {name: 'Steven'}),
@@ -183,7 +176,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at runtime: InvalidArgumentType
 
   Scenario: Floating point parameter for LIMIT with ORDER BY should fail
-    Given any graph
     And having executed:
       """
       CREATE (s:Person {name: 'Steven'}),
@@ -200,7 +192,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at runtime: InvalidArgumentType
 
   Scenario: Floating point LIMIT should fail
-    Given any graph
     And having executed:
       """
       CREATE (s:Person {name: 'Steven'}),
@@ -215,7 +206,6 @@ Feature: SkipLimitAcceptanceTest
     Then a SyntaxError should be raised at compile time: InvalidArgumentType
 
   Scenario: Floating point parameter for SKIP should fail
-    Given any graph
     And having executed:
       """
       CREATE (s:Person {name: 'Steven'}),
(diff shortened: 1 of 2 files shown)


tck/features/SkipLimitAcceptance.feature
```diff
@@ -30,24 +30,24 @@
...
   Scenario: LIMIT with an expression that depends on variables should fail
+    Given any graph
     When executing query:
       """
       MATCH (n) RETURN n LIMIT n.count
       """
     Then a SyntaxError should be raised at compile time: NonConstantExpression
 
   Scenario: SKIP with an expression that does not depend on variables
+    Given an empty graph
```
Mats-SX: I don't see why this can't be `any graph`?

tck/features/SkipLimitAcceptance.feature
```diff
@@ -67,6 +67,7 @@ Feature: SkipLimitAcceptanceTest
 
 
   Scenario: LIMIT with an expression that does not depend on variables
+    Given an empty graph
```
Mats-SX: This should also be possible with `any graph`

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+      | limit | -1 |
+    When executing query:
+      """
+      MATCH (p:Person)
+      RETURN p.name AS name
+      LIMIT $limit
+      """
+    Then a SyntaxError should be raised at runtime: NegativeIntegerArgument
+
+  Scenario: Negative parameter for TOP should fail
```
Mats-SX: What is `TOP`? I would prefer something like
```suggestion
  Scenario: Negative parameter for LIMIT with ORDER BY should fail
```

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+
+  Scenario: Negative parameter for SKIP should fail
+    Given any graph
+    And having executed:
+      """
+      CREATE (s:Person {name: 'Steven'}),
+             (c:Person {name: 'Craig'})
+      """
+    And parameters are:
+      | limit | -1 |
```
Mats-SX: Perhaps let's call this not `limit` but `skip`?

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+      | limit | 1.0 |
+    When executing query:
+      """
+      MATCH (p:Person)
+      RETURN p.name AS name
+      LIMIT $limit
+      """
+    Then a SyntaxError should be raised at runtime: InvalidArgumentType
+
+  Scenario: Floating point parameter for TOP should fail
```
Mats-SX: Again; `TOP` isn't an openCypher term that I know of.

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+
+  Scenario: Floating point parameter for LIMIT should fail
+    Given any graph
+    And having executed:
+      """
+      CREATE (s:Person {name: 'Steven'}),
+             (c:Person {name: 'Craig'})
+      """
+    And parameters are:
+      | limit | 1.0 |
```
Mats-SX: Perhaps let's use a value that doesn't immediately make you think this would actually make sense, such as `1.5`

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+
+  Scenario: Floating point parameter for TOP should fail
+    Given any graph
+    And having executed:
+      """
+      CREATE (s:Person {name: 'Steven'}),
+             (c:Person {name: 'Craig'})
+      """
+    And parameters are:
+      | limit | 1.0 |
```
Mats-SX: Again, the value makes you wonder why this is a bad idea

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+    And having executed:
+      """
+      CREATE (s:Person {name: 'Steven'}),
+             (c:Person {name: 'Craig'})
+      """
+    When executing query:
+      """
+      MATCH (p:Person)
+      RETURN p.name AS name
+      LIMIT 1.0
```
Mats-SX: Value

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+      | limit | 1.0 |
+    When executing query:
+      """
+      MATCH (p:Person)
+      RETURN p.name AS name
+      ORDER BY name LIMIT $limit
+      """
+    Then a SyntaxError should be raised at runtime: InvalidArgumentType
+
+  Scenario: Floating point LIMIT should fail with a syntax exception
```
Mats-SX: Let's not state details of the assertion in the name
```suggestion
  Scenario: Floating point LIMIT should fail
```

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+      | limit | -1 |
+    When executing query:
+      """
+      MATCH (p:Person)
+      RETURN p.name AS name
+      ORDER BY name LIMIT $limit
+      """
+    Then a SyntaxError should be raised at runtime: NegativeIntegerArgument
+
+  Scenario: Negative LIMIT should fail with a syntax exception
```
Mats-SX: Let's not state details of the assertion in the name

 ```suggestion
  Scenario: Negative LIMIT should fail 
```

File: tck/features/SkipLimitAcceptance.feature [hunk 11]
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+      | limit | -1 |
+    When executing query:
+      """
+      MATCH (p:Person)
+      RETURN p.name AS name
+      SKIP $limit
+      """
+    Then a SyntaxError should be raised at runtime: NegativeIntegerArgument
+
+  Scenario: Negative SKIP should fail with a syntax exception
```
Mats-SX: Let's not state details of the assertion in the name

```suggestion
  Scenario: Negative SKIP should fail 
```

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+
+  Scenario: Floating point parameter for SKIP should fail
+    Given any graph
+    And having executed:
+      """
+      CREATE (s:Person {name: 'Steven'}),
+             (c:Person {name: 'Craig'})
+      """
+    And parameters are:
+      | limit | 1.0 |
```
Mats-SX: `3.14` or some more floaty value preferred

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+    When executing query:
+      """
+      MATCH (p:Person)
+      RETURN p.name AS name
+      SKIP $limit
+      """
+    Then a SyntaxError should be raised at runtime: InvalidArgumentType
+
+  Scenario: Floating point SKIP should fail with a syntax exception
```
Mats-SX:
```suggestion
  Scenario: Floating point SKIP should fail
```

tck/features/SkipLimitAcceptance.feature
```diff
@@ -82,3 +83,165 @@ Feature: SkipLimitAcceptanceTest
...
+    And having executed:
+      """
+      CREATE (s:Person {name: 'Steven'}),
+             (c:Person {name: 'Craig'})
+      """
+    When executing query:
+      """
+      MATCH (p:Person)
+      RETURN p.name AS name
+      SKIP 1.0
```
Mats-SX: 2.7

tck/features/SkipLimitAcceptance.feature
sherfert: Since this one only returns whether the count is greater than 0, I agree.

tck/features/SkipLimitAcceptance.feature
sherfert: The assertion that we get a count of 2 does not hold for `any graph`.

Mats-SX: I think it does hold, since we create 3 nodes in the init step, then match for any nodes and limit to 2, the count is always going to be 2, regardless of how many nodes were in the graph, or what shape they had.

The only way this would not hold is if we were to take constraints into account. However, constraints are still considered out of scope for the TCK (and openCypher in general), so this is a non-argument. And either way, if openCypher adopted constraints the way they are implemented in Neo4j, the creations in the init step would still be possible because there is no label on the nodes.

So I insist that `any graph` is fine :)

sherfert: You are right.

PR #383, opened 2019-09-12
Bnfg4 to o c xml

PeterFurniss:
xml schema for grammar xml;  convert to/from xml, sql-bnf, antlr4 g4


tools/grammar/src/main/java/org/opencypher/tools/grammar/TranslateGrammar.java
```diff
@@ -0,0 +1,152 @@
...
+        } else if ( inG4 != null) {
+            inputOptions++;
+            G4Processor g4Processor = new G4Processor();
+            grammar = g4Processor.processFile(inG4);
+        } else if (  inBnf != null) {
+            inputOptions++;
+            BNFProcessor bnfProcessor = new BNFProcessor();
+            grammar = bnfProcessor.processFile(inBnf);
+        }
+        if (inputOptions != 1) {
```
thobe: This check doesn't actually work. The increments to `inputOptions` are performed in mutually exclusive code blocks (`else if`).

PR #384, opened 2019-10-08
Stop publishing snapshots to retiring m2.neo4j.org

Mats-SX:
We upload snapshot artifacts to S3 and there is no need to retain
module build snapshots on a public server for every check-in. Local
builds should cover this need, coupled with Maven Central releases for
stable versions.


thobe:
This change itself seems like a very good thing.

I would like to know if there is still a need to publish executable artefacts _somewhere_?
Because if there is we should find a new place to publish to.

Mats-SX:
@thobe For snapshots I don't see that need as a very great one. We publish stable versions as part of our milestone releases and snapshots can be built locally by someone who needs it. Our CI builds make sure that snapshot building is always possible.

PR #408, opened 2020-04-16
Tag to switch off code style checker – cleaned up history

hvub:
This PR adapts FeatureFormatChecker to not check code style if a scenario has a @withIntentionalStyleViolation tag. This allow to write scenarios with intentional code style violations (or work around limitation of the code style checker).

This PR replaces [PR 405](https://github.com/opencypher/openCypher/pull/405), which had a messed-up history.

### DIFF
--- a/tck/features/uncategorized/ColumnNameAcceptance.feature
+++ b/tck/features/uncategorized/ColumnNameAcceptance.feature
@@ -58,6 +58,7 @@ Feature: ColumnNameAcceptance
       | nOdEs( p ) |
     And no side effects
 
+  @withIntentionalStyleViolation
   Scenario: Keeping used expression 3
     When executing query:
       """
--- a/tck/features/uncategorized/ColumnNameAcceptance.feature
+++ b/tck/features/uncategorized/ColumnNameAcceptance.feature
@@ -58,7 +58,7 @@ Feature: ColumnNameAcceptance
       | nOdEs( p ) |
     And no side effects
 
-  @withIntentionalStyleViolation
+  @skipStyleCheck
   Scenario: Keeping used expression 3
     When executing query:
       """
--- a/tools/tck-integrity-tests/src/main/scala/org/opencypher/tools/tck/FeatureFormatChecker.scala
+++ b/tools/tck-integrity-tests/src/main/scala/org/opencypher/tools/tck/FeatureFormatChecker.scala
@@ -39,10 +39,18 @@ import scala.util.Try
 
 class FeatureFormatChecker extends TCKCucumberTemplate {
 
+  private var currentScenarioName = ""
+  private var withIntentionalStyleViolation: Boolean = false
+
+  Before { (scenario: cucumber.api.Scenario) =>
+    validateDuplicateNames(scenario).map(msg => throw InvalidFeatureFormatException(msg))
+    currentScenarioName = scenario.getName
+    withIntentionalStyleViolation = scenario.getSourceTagNames.contains("@withIntentionalStyleViolation")
+  }
+
   private var lastSeenQuery = ""
   private val orderBy = "(?si).*ORDER BY.*"
   private val call = "(?si).*CALL.*"
-  private var currentScenarioName = ""
   private val stepValidator = new ScenarioFormatValidator
 
   Background(BACKGROUND) {}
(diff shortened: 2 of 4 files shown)


tools/tck-integrity-tests/src/main/scala/org/opencypher/tools/tck/FeatureFormatChecker.scala
```diff
@@ -160,10 +163,6 @@ class FeatureFormatChecker extends TCKCucumberTemplate {
     }
   }
 
-  private val scenariosWithIntentionalStyleViolations =
-    Set("Keeping used expression 3",
-        "Keeping used expression 4")
-
```
Mats-SX, line 166: Nice improvement moving these out to the scenarios themselves. Also improves understandability of scenarios

tools/tck-integrity-tests/src/main/scala/org/opencypher/tools/tck/FeatureFormatChecker.scala
```diff
@@ -39,10 +39,18 @@ import scala.util.Try
 
 class FeatureFormatChecker extends TCKCucumberTemplate {
 
+  private var currentScenarioName = ""
+  private var withIntentionalStyleViolation: Boolean = false
+
+  Before { (scenario: cucumber.api.Scenario) =>
+    validateDuplicateNames(scenario).map(msg => throw InvalidFeatureFormatException(msg))
+    currentScenarioName = scenario.getName
+    withIntentionalStyleViolation = scenario.getSourceTagNames.contains("@withIntentionalStyleViolation")
```
Mats-SX, line 48: I would move `@withIntentionalStyleViolation` to a constant. Perhaps in `TCKConstants` class?

tck/features/uncategorized/ColumnNameAcceptance.feature
```diff
@@ -58,6 +58,7 @@ Feature: ColumnNameAcceptance
       | nOdEs( p ) |
     And no side effects
 
+  @withIntentionalStyleViolation
```
Mats-SX, line 61: I think the tag name is a bit long. I suggest some alternatives:

`@BadStyle`
`@StyleViolation`

I think the presence of the tag indicates the 'intentional' part to the reader and doesn't have to be included.

Or taking a different approach, more towards the effect of the tag rather than its definition:

`@NoStyleCheck`
`@SkipStyleCheck`

tools/tck-integrity-tests/src/main/scala/org/opencypher/tools/tck/FeatureFormatChecker.scala
```diff
@@ -39,10 +39,18 @@ import scala.util.Try
 
 class FeatureFormatChecker extends TCKCucumberTemplate {
 
+  private var currentScenarioName = ""
+  private var withIntentionalStyleViolation: Boolean = false
```
Mats-SX, line 43: Also here something shorter like `skipStyleCheck` could be used.

tck/features/uncategorized/ColumnNameAcceptance.feature
hvub: I do not have strong opinion and change the tag name like three times before opening the PR. I agree it is fairly long. Out of you proposals I lean towards `@skipStyleCheck`. That is uses instructional language, like `@ignore`, and I can see a rule saying: 

- Instructional language for tags that make the API, etc. behave differently 
- Non-instructional language for tags that mark language feature used in a scenario (something, we do not yet have, but anticipate).

Concluding, I will change the tag to `@skipStyleCheck` and move it to a constant, as well. (Should do the same with `@ignore`, which I remember seen being use as string literal directly in code.)

Mats-SX: Sounds good. Personally I would go for an upper-case first letter, also for `@Ignore` but I'll concede that if you disagree.

Mats-SX: If you would like to keep the lower-case, perhaps a different scheme than camel case? Like `@skip_style_check` or `@skip-style-check`?

hvub: I do not have a particular opinion other than that the existing tags @ignore and @allowCustomError start with a lower case letter. @ignore may also be used in custom tests outside the TCK but through the TCK API. Hence, changing this is sort of a breaking API change. We could allow also lower case for the old tags to not break the API and still changing to upper case. Is it worth the effort?

Mats-SX: No probably not. Consistency is probably more important here than stylism. Let's go with `@skipStyleCheck`. I'm coming around to it :)

PR #411, opened 2020-04-21
map operations

DvirDukhan:
@hvub 
Map operations PR
All scenarios in `Map2.feature` and `Map4.feature` are extracted from `ExpressionAcceptance.feature` and they are all added with `Given any graph` as `ExpressionAcceptance.feature` had 
```
  Background:
    Given any graph
```

I'll add scenario numbering once the file structure and content is approved.
Thanks


hvub:
A couple of questions:

1. What do you expect Map1.feature and Map3.feature to be?
2. Three scenarios have been removed according to the diff. Is that by mistake or what is the motivation for it?
   * [⟩ uncategorized ⟩⟩ ExpressionAcceptance - Fail at runtime when attempting to index with an Int into a Map](https://github.com/opencypher/openCypher/pull/411/files#diff-fcb37079777f5c6a21534ff1534ea826L107)
   * [⟩ uncategorized ⟩⟩ ExpressionAcceptance - Fail at runtime when trying to index into a map with a non-string](https://github.com/opencypher/openCypher/pull/411/files#diff-fcb37079777f5c6a21534ff1534ea826L118)
   * [⟩ uncategorized ⟩⟩ ExpressionAcceptance - Fail at runtime when trying to index something which is not a map or list](https://github.com/opencypher/openCypher/pull/411/files#diff-fcb37079777f5c6a21534ff1534ea826L129)

Thanks.

hvub:
Hmm ... scrolling over "Files changed" again this morning, I wonder what I actually looked at yesterday. Let me have a second look.

hvub:
Okay, I had a second look. Please, ignore my comments 1 and 2 from above.

However, please rename `Map4.Feature` to `Map4.feature` (lower case `f`). The API parses only `.feature` files. Hence, Map4 scenarios did not show up in the diff.

Further, I think we should 

- either merging static and dynamic element access for maps.
- or splitting element access for list into static and dynamic as well

so that list and map categorization is aligned in that respect.

I understand the distinction between static and dynamic element access, as static is any access where the element to access can be figure out at compile time, i.e. without accessing other data or any parameters. Dynamic are all element accesses that are not static. In that sense, Map2 "Execute n['name'] in read queries" and "Execute n['name'] in update queries" as well as List1 "Indexing into literal list" and "Indexing into nested literal lists" are static.

Since dynamic list access is harder to implement particularly if an implementation uses static type checking, I suggest we indeed separate static and dynamic element accesses for lists and maps.

Concretely, I suggest that you 
- Move Map2 "Execute n['name'] in read queries" and "Execute n['name'] in update queries" to Map1 
- Introduce a new List2, renumbering all subsequent ListX files, move List1 "Use list lookup based on parameters when there is lhs type information", "Use list lookup based on parameters when there is no type information" and "Use list lookup based on parameters when there is rhs type information" to that new List2.
- Align the feature names of Map1 and Map2 with List1 and List2 to "...1 - Static Element Access" and  "...2 - Dynamic Element Access", respectively.

Further, I think you can all do this in this PR. If you have commit messages on ListX saying something like "Aligning with map element access", it will look reasonable.

Does that make sense?

Thanks. -H

tck/features/expressions/map/Map1.feature
```diff
@@ -0,0 +1,31 @@
...
+# should not be described as “Cypher” (and Cypher® is a registered trademark of
+# Neo4j Inc.) or as "openCypher". Extensions by implementers or prototypes or
+# proposals for change that have been documented or implemented should only be
+# described as "implementation extensions to Cypher" or as "proposed changes to
+# Cypher that are not yet approved by the openCypher community".
+#
+
+#encoding: utf-8
+
+Feature: Map1 - Static Value Access
```
hvub, line 31:
```suggestion
Feature: Map1 - Static Value Access
# Static value access refers to the dot-operator – <expression resulting in a map>.<identify> – which does not allow any dynamic computation of the map key – i.e. <identify>.
```

Please add a comment clarifying the understanding of the term Static value access.

tck/features/expressions/map/Map2.feature
```diff
@@ -28,12 +28,10 @@
 
 #encoding: utf-8
 
-Feature: ExpressionAcceptance
-
-  Background:
-    Given any graph
+Feature: Map2 - Dynamic Value Access
```
hvub, line 31:
```suggestion
Feature: Map2 - Dynamic Value Access
# Dynamic value access refers to the bracket-operator – <expression resulting in a map>'['<expression resulting in a string>']' – irrespectively of whether the map key – i.e. <expression resulting in a string> – could be evaluated statically in a given scenario.
```

Please add a comment clarifying the understanding of the term Dynamic value access.

PR #414, opened 2020-04-24
Categorize SET

hvub:
This PR categorizes scenarios for the SET clause.


petraselmer:
LGTM.

I think the more purposeful naming of some scenarios is an improvement, as is the filling-in of hitherto missing scenarios.

I had wondered whether Set 4 and 5 couldn't be merged - as they both deal with setting multiple properties - but on second thought, it may not be a bad idea to keep them separate for now. Depending on the extent to which the full complement of missing scenarios are identified and filled in at a later stage will dictate as to whether these two files should indeed become one. 

This PR can be merged

hvub:
Travis indicates that CI test ran and finished green:

https://travis-ci.org/github/opencypher/openCypher/builds/679093636

Github just did not learn about it.

I will merge. Thanks.

PR #418, opened 2020-05-18
Generator for tck/index.adoc

hvub:
The [tck/index.adoc](tck/index.adoc) is intended as repository documentation for the feature categories.

This PR adds [GenerateTCKIndexDoc](https://github.com/opencypher/openCypher/compare/master...hvub:create-tck-index-during-build?expand=1#diff-715e7bfab98083a2e2eda02bffafb292R44), which generates the tck/index.adoc based on folder structure and .feature files. It excludes the `uncategorized` and `precategorized` folders. The generation is fairly simple at the moment. However, it keeps tck/index.adoc uptodate, which no categorization step so far did. If it proofs to be useful, it is possible to improve the generation.

The PR also add [GenerateTCKIndexDocTest](https://github.com/opencypher/openCypher/compare/master...hvub:create-tck-index-during-build?expand=1#diff-715e7bfab98083a2e2eda02bffafb292R39), which runs GenerateTCKIndexDoc as a test, so it run during build – without testing anything in particular.

Keeping a generated artifact in the repository is not ideal, of course. It seems reasonable to remove the GenerateTCKIndexDoc from the build ones the categories become stable or even earlier if the generation shows to cause too many merge conflicts or other troubles.

### DIFF
--- a/tck/index.adoc
+++ b/tck/index.adoc
@@ -8,10 +8,10 @@ Within each member, there are additional categories.
 There is also an `uncategorized` and `precategorized` directory containing uncategorized features.
 
 
-== clauses
+== Clauses
 
 
-=== create
+=== Create
 
 * Create1 - Creating nodes
 * Create2 - Creating relationships
@@ -20,7 +20,7 @@ There is also an `uncategorized` and `precategorized` directory containing uncat
 * Create5 - Multiple hops create patterns
 * Create6 - Negative tests
 
-=== delete
+=== Delete
 
 * Delete1 - Deleting nodes
 * Delete2 - Deleting relationships
@@ -29,7 +29,7 @@ There is also an `uncategorized` and `precategorized` directory containing uncat
 * Delete5 - Delete clause interoperation with built-in data types
 * Delete6 - Negative scenarios
 
-=== match
+=== Match
 
 * Match1 - Match Nodes scenarios
 * Match10 - Match clause failure scenarios
@@ -42,7 +42,12 @@ There is also an `uncategorized` and `precategorized` directory containing uncat
 * Match8 - Match clause Interoperation with other clauses
 * Match9 - Match deprecated scenarios
 
-=== set
+=== Remove
+
+* Remove1 - Remove a Property
+* Remove2 - Remove a Label
+
+=== Set
 
 * Set1 - Set a Property
 * Set2 - Set a Property to Null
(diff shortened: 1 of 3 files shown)


Mats-SX:
This seems like a good idea. I think one option to make sure this is kept up to date without having to think about how to have a generated file be checked in could be to adapt the test so that the test actually tests that the checked-in file corresponds to what is being generated. Essentially:

generate the file
read the file
compare contents
fail if there's a difference

An advantage to this approach is that changes to the TCK will not be accepted on PR builds unless the index is also updated. This forces updating the index in conjunction with the contents, which can be beneficial to `git blame` where one can follow the evolution of the TCK via the changes made to the index, as well as the content.

hvub:
@Mats-SX I tried it out, see above. The problem that arise from this setup is that it forces every PR to be rebased on the latest master, generate the index.adoc, otherwise CI fails. If another TCK PR gets merged in between, your PR has to go through this routine again. Clearly it ensures the index.adoc is always up-to-date. On the downside, it complicates the dev processes. It seems worth ones TCK categorizes have settled more. At the moment, I am ensure if it is worth the hassle it causes.

Mats-SX:
I don't understand how that can be. This PR only contains changes made on this PR. How is it being influenced by not being in sync with master? The generated index should only consider the state of this PR, and the test generates it using the generator. If the test fails it indicates that changes made on _this_ PR causes the index to be incorrect.

Perhaps the CI test doesn't check out the PR branch and runs on just that? But instead it merges this PR onto latest master and then runs the test? If that could be changed then the test wouldn't get in your way.

Mats-SX:
Ah, indeed, that's what it does:

```
git fetch origin +refs/pull/418/merge
```

I think the `/merge` part here checks out this PR as merged into master. If you can change that you can retain the test.

But I'll let you make a call on that since I'm not really working on this project. Before I realised that the above could be the case I wrote another comment, and I'll reference it here. Feel free to ignore:

I would say that in a perfect world, merges should only happen for PRs that are based directly on `master`, because merging out-of-sync PRs is a risk to `master` stability as the PR check is not 100% reliable (it does not consider the changes made by other PRs merged since last rebase). However, in real life 100% reliability is not always necessary, so in our team we make compromises based on best-effort estimation of the actual risk, influenced by the rate at which PRs come in and set all other PRs out of sync. I would tend to think that rebasing TCK category PRs would be fairly quick and conflict-free, but I am not the one doing it so I don't actually know and I wouldn't want to suggest you do more work than you think is reasonable.

hvub:
I commented out the verification test. At the moment, the TCK category structure is very dynamic, hence

- It is not super important the index.adoc is always on top of things. If index.adoc falls behind due to concurrent PRs it will get updated with next wave of PRs. 
- There is enough value in having  it automatically updated and getting eventual consistent.
- We can easily reactivate the verification test once the TCK category structure gets more stable.

PR #419, opened 2020-05-18
Categorizes temporal functions

hvub:
This PR categorizes temporal functions. It also factors in all backgrounds. Otherwise, no scenarios have been changed.

Note that Temporal4 contains scenarios that test storage capabilities for temporal values. These could also be categorized under the CREATE clause. However, given that temporal may be considered as advanced functionality, it seems sensible to bundle all its aspects in one category.

### DIFF
--- a/tck/features/expressions/temporal/Temporal3.feature
+++ b/tck/features/expressions/temporal/Temporal3.feature
@@ -27,7 +27,7 @@
 #
 
 #encoding: utf-8
-Feature: Temporal3 - Create Temporal Values from other Temporal Values
+Feature: Temporal3 - Project Temporal Values from other Temporal Values
 
   Scenario Outline: [1] Should select date
     Given any graph
--- a/tck/features/expressions/temporal/Temporal3.feature
+++ b/tck/features/expressions/temporal/Temporal3.feature
@@ -27,12 +27,10 @@
 #
 
 #encoding: utf-8
-Feature: TemporalSelectAcceptance
+Feature: Temporal3 - Create Temporal Values from other Temporal Values
 
-  Background:
+  Scenario Outline: [1] Should select date
     Given any graph
-
-  Scenario Outline: Should select date
     When executing query:
       """
       WITH <other> AS other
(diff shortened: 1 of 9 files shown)


tck/features/expressions/temporal/Temporal3.feature
```diff
@@ -27,12 +27,10 @@
 #
 
 #encoding: utf-8
-Feature: TemporalSelectAcceptance
+Feature: Temporal3 - Create Temporal Values from other Temporal Values
```
thobe, line 30: I would call this "_Selecting_ Temporal Values from other Temporal Values"

tck/features/expressions/temporal/Temporal5.feature
```diff
@@ -28,12 +28,10 @@
 
 #encoding: utf-8
 
-Feature: TemporalAccessorAcceptance
+Feature: Temporal5 - Access Parts of Temporal Values
```
thobe, line 31: I think we have called these _Parts_ "_Components_": https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-08-06-date-time.adoc#accessing-components-of-temporal-values

So I'd call this feature "Accessing Components of Temporal Values".

hvub:
```suggestion
Feature: Temporal5 - Access Components of Temporal Values
```

Yes, great suggestion. Will adopt. Thx.

tck/features/expressions/temporal/Temporal3.feature
hvub: I thought about that. The original file also the term _selection_. Coming from the natural meaning of the word [_select_](https://www.merriam-webster.com/dictionary/select), the operation portraited in these scenarios allows to do more than just choosing components* to get a new temporal value. It allow to do overwrite components, too. If you wanted to refer to the relational projection, which I can see this operation has similarity too, I would not detour via SQL's SELECT, but go straight to it and speak of `Projecting Temporal Values from other Temporal Values`.

[*] As a side note, while I skimmed these scenarios, I wondered if `datetime`, `date`, and `time` are the only components one can select or if you can also do, say `{hour: other, minute: other}`. Probably not, there are no scenarios for that and `hour`, etc. are expected to be integers, I guess.

hvub:
```suggestion
Feature: Temporal3 - Project Temporal Values from other Temporal Values
```

hvub: Just saw in [the CIP]( https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-08-06-date-time.adoc#247-converting-and-composing-temporal-instant-values) and tried in the product that you can select `hour`, etc. So it is a gap in the scenarios. I will create a card about that.

hvub: _Compose_ is another verb that is can be used here and is also used by the [CIP](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-08-06-date-time.adoc#247-converting-and-composing-temporal-instant-values).

```suggestion
Feature: Temporal3 - Compose Temporal Values from other Temporal Values
```

PR #428, opened 2020-07-12
elemntary graph operations

DvirDukhan:
@hvub 
I think we should iterate this one.
Added graph operations and expressions
also, changes to existing files:
`LabelsAcceptance.feature` - left with only negative tests - Add `@NegativeTest` annotation and added `Given` to each scenario instead of background
`Map2.feature`, `Map3.feature` - Left with scenarios on explicit maps
`NullAcceptance` moved to `Graph9.feature


hvub:
Some remarks

 * `Graph8 - Compute the number of subgraphs induced by a pattern expression` is an odd one. I understand it comes from the spreadsheet. The problem probably is Cypher's overloading of the function `size()`. There is (1) `size(<list>)` ([Manual](https://neo4j.com/docs/cypher-manual/current/functions/scalar/#functions-size)) and (2) `size(<pattern>)` ([Manual](https://neo4j.com/docs/cypher-manual/current/functions/scalar/#functions-size-of-pattern-expression)). Variant 2 is the one meant here. The scenario `Functions should return null if they get path containing unbound` is about the Variant 1, so it does not really fit here and we will have to find another place for it. In the meantime it could be simply be moved back or into some `precategorized` feature.
 * Thinking about the order (and naming), my proposal is:
    1. Node and edge identifier - ID function
    2. Edge source and destination
    3. Node labels
    4. Edge relationship type
    5. Static property access
    6. Dynamic property access
    7. Properties keys  //nodes and edges
    8. Check if a property exists
    9. Properties as maps
    10. Compute the number of subgraphs induced by a pattern expression

Others it looks good at the first glance.

hvub:
Let's remove `Compute the number of subgraphs induced by a pattern expression` from this category. It belong into something like `Patter operations`.

tck/features/uncategorized/FunctionsAcceptance.feature
```diff
@@ -30,49 +30,21 @@
...
-      OPTIONAL MATCH p = (a)-[r]->()
-      RETURN size(nodes(p)), type(r), nodes(p), relationships(p)
-      """
-    Then the result should be, in any order:
-      | size(nodes(p)) | type(r) | nodes(p) | relationships(p) |
-      | null           | null    | null     | null             |
-    And no side effects
-
-  Scenario: `properties()` on a node
+  Scenario: Run coalesce
```
hvub, line 33: The scenario "Run coalesce" should be in "Conditional1 - Coalesce Expression" by commit https://github.com/RedisGraph/openCypher/commit/35f7e54042eadeac33c5986c91cf1e0b1a528a70

It seems from https://github.com/RedisGraph/openCypher/commits/elemntary_graph_operations that you have rebase the our branch, however it still wants to duplicate this scenario back into uncategorized/FunctionsAcceptance. Please double check.

tck/features/uncategorized/FunctionsAcceptance.feature
```diff
@@ -121,24 +93,24 @@ Feature: FunctionsAcceptance
...
     And no side effects
 
-  Scenario: `exists()` with dynamic property lookup
-    Given an empty graph
-    And having executed:
-      """
-      CREATE (:Person {name: 'foo'}),
-             (:Person)
-      """
+  Scenario Outline: `exists()` with literal maps
```
hvub, line 96: The scenario outline "`exists()` with dynamic property lookup" should be in "Map4 - Exists function" by commit https://github.com/RedisGraph/openCypher/commit/4ae580e8880575d63d0daec6114e4446edc4894b

It seems from https://github.com/RedisGraph/openCypher/commits/elemntary_graph_operations that you have rebase the our branch, however it still wants to duplicate this scenario back into uncategorized/FunctionsAcceptance. Please double check.

tck/features/precategorized/operations/patterns/Pattern1.feature
```diff
@@ -0,0 +1,44 @@
...
+# proposals for change that have been documented or implemented should only be
+# described as "implementation extensions to Cypher" or as "proposed changes to
+# Cypher that are not yet approved by the openCypher community".
+#
+
+#encoding: utf-8
+
+Feature: Pattern1 - Compute the number of subgraphs induced by a pattern expression
+
+  Scenario: Functions should return null if they get path containing unbound
```
hvub, line 33: This is a very weird scenarios. Hard to place. Probably need to be rip up. 

I like putting something in `precategorized` to not through away work. However, `Pattern1` look wrong. Two of the functions are path functions (`nodes`, `relationships`), one is a graph function (`types`), and one is a list functions (`size`), non is a pattern function. So, I guess, I would leave this scenario in `uncategorized/FunctionsAcceptance.feature` for now and deal with it in the final clean up round of categorization, when we will probably rip it in four pieces. 

You could add a comment to the scenario like "Consider splitting into one scenario per one of four function:`nodes`, `relationships`, `types`, and `size`." to keep record of this intension.

hvub (changes requested):
It looks pretty good now, with only a few rough spots on the edges.

Please, address the four comments and add the scenario numbering. If that is done, I expect, it will be ready to merge.

Thanks.

tck/features/uncategorized/FunctionsAcceptance.feature
DvirDukhan: removed

tck/features/uncategorized/FunctionsAcceptance.feature
DvirDukhan: removed

tck/features/precategorized/operations/patterns/Pattern1.feature
DvirDukhan: restored to `uncategorized/FunctionsAcceptance.feature`. Comment added

hvub:
Great. Please add the scenario numbering and then it is good to go in.

Thanks. -H

PR #431, opened 2020-07-30
Missing node variable in test

AviAvni:

PR #433, opened 2020-07-31
Add missing node variable

Mats-SX:
Thanks to @AviAvni for raising this issue in #431


thobe:
From the look of this test, it seems we are not continuously verifying the correctness of the TCK.

@hvub I have some memory of you looking into setting up some such continuous verification using Neo4j, what is the status of that?

thobe:
This particular issue would have been caught by running the semantic analysis available in the https://github.com/opencypher/front-end

PR #439, opened 2020-08-19
Path operations

DvirDukhan:
@hvub 
Path operations PR
1. Path1 and Path3 scenarios are duplicated (see commit messages) as they are not purely testing the functions themselves
2. Path2 scenarios moved from match 6-4


tck/features/expressions/path/Path3.feature
```diff
@@ -0,0 +1,82 @@
...
+      """
+      MATCH p = (n)-->(x)
+      WHERE length(p) = 10
+      RETURN x
+      """
+    Then the result should be, in any order:
+      | x |
+    And no side effects
+
+  Scenario: Return a var length path of length zero
```
hvub, line 66: I think this does not have to be duplicated. It is clearly testing the length function.

Hence, it should only be here (removed from its original feature) and be at the top of this feature file.

tck/features/expressions/path/Path3.feature
```diff
@@ -0,0 +1,82 @@
...
+# proposals for change that have been documented or implemented should only be
+# described as "implementation extensions to Cypher" or as "proposed changes to
+# Cypher that are not yet approved by the openCypher community".
+#
+
+#encoding: utf-8
+
+Feature: Path3 - Length of a path
+
+  Scenario: Pass the path length test
```
hvub, line 33: I am torn on this one and the next one. I see where you coming from, but I like to avoid duplication as much as possible. The problem is, we have not yet a clear idea how to categorize `WHERE`. Right now I am thinking, we could categorize `WHERE` into stuff like filtering on atomic values, compound values, node characteristics, relationship characteristics, path characteristics, aggregates, complex predicates, etc. 

In that case, these scenarios should go into `WHERE` filtering on path characteristics.

tck/features/expressions/path/Path1.feature
```diff
@@ -0,0 +1,76 @@
...
+# proposals for change that have been documented or implemented should only be
+# described as "implementation extensions to Cypher" or as "proposed changes to
+# Cypher that are not yet approved by the openCypher community".
+#
+
+#encoding: utf-8
+
+Feature: Path1 - Nodes of a path
+
+  Scenario: Aggregation of named paths
```
hvub, line 33: Similarly to my rationale given on Path3, I think, I would rather have these two scenarios in some aggregation feature and leave these feature empty for know.

hvub (changes requested):
I principle it looks good, but I have reservation about duplication. I think, I prefer empty feature other duplication. Empty feature scream more for being filled with properly design scenarios, than features filled with ill-fitted duplicates scream for being fixed.

Hence, I suggest to roll back the duplication in one way of the other, cf. inline comments.

tck/features/expressions/path/Path3.feature
DvirDukhan: removed
only on match6-2

tck/features/expressions/path/Path1.feature
DvirDukhan: removed

hvub (changes requested):
Looks good. Please add the scenario numbering and then it is good to go.

PR #105, opened 2016-06-29
Call procedures

boggle:
[CIP2015-06-24](https://github.com/boggle/openCypher/blob/CIP2015-06-24-call-procedures/cip/1.accepted/CIP2015-06-24-call-procedures.adoc)

PR #449, opened 2020-09-29
Call clause PR

DvirDukhan:
@hvub 
1. I split the scenarios according to their procedure input type. I wonder if this resolution is too fine.
1. All scenarios now start with `Given an empty graph` instead of global background.
1. renamed `Standalone call to procedure that takes no arguments` to `Standalone call to STRING procedure that takes no arguments`
1. renamed `In-query call to procedure that takes no arguments` to `In-query call to STRING procedure that takes no arguments`

As always, I will add the numbering once we decide the final structure of this PR.
Thanks


DvirDukhan:
change list
- [ ] in query call
- [ ] standalone call 
- [ ] single parameter
- [ ] multiple parameters

hvub:
Hi @DvirDukhan 

I have given this a bit more detailed thinking and suggest the following:

 * `Call1 - Basic procedure calling`
   * Scenarios that test that a procedure is called and that it is called the right number of times and its results is passed to the caller
   * Mainly what is Call1 currently
   * That a procedure is called the right number of times, seems not really tested at the moment
 * `Call2 - Same-type arguments`
   * Scenarios that test that procedure arguments are passed to parameters, where the arguments are of the same or an incompatible type as the corresponding parameters
   * Mainly what is Call3, Call4, Call5 currently
 * `Call3 - Assignable-type arguments`
   * Scenarios that test that procedure arguments are passed to parameters, where the arguments are not of the same type as the corresponding parameters but of an assignable type 
   * What is Call2 currently
 * `Call4 - Null arguments`
   * Scenarios that test how a procedure handles null argument(s)
   * What is Call6 currently
 * `Call5 - Results projection`
   * Scenarios that test YIELD behavior on top of plainly passing through result, such as yielding nothing, yielding a subset of columns, yield renamed column

All negative test scenarios should be placed into one of these features.

The current scenarios do not perfectly fit this categorization. Let's place them where they fit best. We can massage them later to fit better.

Cheers. -H

DvirDukhan:
@hvub 
fixed PR comments

tck/features/clauses/call/Call1.feature
```diff
@@ -0,0 +1,280 @@
...
+    Then the result should be empty
+    And no side effects
+
+  Scenario: In-query call to procedure that takes no arguments and yields no results
+    Given an empty graph
+    And there exists a procedure test.doNothing() :: ():
+      |
+    When executing query:
+      """
+      CALL test.doNothing() YIELD - RETURN 1
```
AviAvni, line 120: why this "YIELD -" is not lligal using the antlr parser?

thobe: I agree, I don't believe `-` should be a valid item after `YIELD`. I don't think that syntax was ever approved.
The specification ([CIP2015-06-24](https://github.com/openCypher/openCypher/blob/master/cip/1.accepted/CIP2015-06-24-call-procedures.adoc)) does not say anything about `YIELD -`, but the pull request that introduced that spec (#105) contains these test cases and a corresponding change to the grammar.
There seems to have been some inconsistencies, @boggle

hvub: Thanks for pointing this out. We have to investigate this to see where this inconsistency originates from.

hvub: After investigation, it was found that this is indeed a bug in the TCK and we should remove the scenario. I will make a suggestion to the PR in that regard.

tck/features/clauses/call/Call1.feature
```diff
@@ -0,0 +1,280 @@
...
+      """
+    Then the result should be, in any order:
+      | 1 |
+    And no side effects
+
```
hvub, line 125:
```suggestion
```

`YIELD -` is not an openCypher feature, so let's remove the scenario.

hvub (changes requested):
@DvirDukhan this is the change necessary to addressing @AviAvni's comment. Thanks.

tck/features/clauses/call/Call5.feature
```diff
@@ -0,0 +1,52 @@
...
+    When executing query:
+      """
+      WITH 'Stefan' AS name, 1 AS id
+      CALL test.my.proc(name, id) YIELD -
+      RETURN name, id, count(*) AS count
+      """
+    Then the result should be, in order:
+      | name     | id | count |
+      | 'Stefan' | 1  | 1     |
+    And no side effects
```
hvub, line 52:
```suggestion
```

This also needs to be remove for the same `YIELD -` reason.

hvub (changes requested):
One more `YIELD -` one to remove. Sorry for missing it in the previous review.

hvub (approved):

PR #452, opened 2020-10-08
Updating versions of various dependencies

hvub:
This PR address the [vulnerability alter on the log4j dependency](https://github.com/opencypher/openCypher/network/alert/tools/grammar/pom.xml/log4j:log4j/open). 

It replaces SLF4j and Log4j 1.2.17 with [log4j-slf4j-impl 2.13.3](https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/index.html). It also migrates the log4j.xml configuration file to Log4j 2 according to the [migration documentation](https://logging.apache.org/log4j/2.x/manual/migration.html).

Further, the PR updates the versions of dependencies as follows:

| Dependency | Old version | New version | Comment |
|---|---|---|---|
| jUnit 5 | 5.4 | 5.7 |
| Cucumber | 4.2.0 | 5.7.0 | It will be involved to go to 6.x, due to Gherkin syntax changes and even more API refactoring |
| Hamcrest | 1.3 | 2.2 |
| Maven Surefire Plugin | 2.22.1 | 2.22.2 |
| Maven Source plugin | 3.0.0 | 3.2.1 |
| Commons CSV | 1.6 | 1.8 |
| Java | 1.8 | 11 |
| Freemarker | 2.3.23 | 2.3.30 |
| Batik | 1.8 | 1.13 |
| XMLgraphics Commons | 2.0 | 2.4 |
| Jackson Dataformat XML | 2.9.0 | 2.11.3 |

The PR also remove the following dependencies, which are not actually used:

 * ScalaCheck
 * APIguardian
 * jUnit 4

Railroad generation fails on some machines, since the default locale causes the generator to generate decimal number with commas (instead of dots) in the SVG attributes. To fix this, this PR modifies the `upload-release.sh` script to explicitly set the default locale to `en_US` when calling the railroad generator with the [parameters `--user.country=US --user.language=en`](https://github.com/opencypher/openCypher/pull/452/files#diff-1c45d9fe570a689062aef6da5bde137cR29-R30), which the `launch.sh` script translates into the respective [JVM parameters](https://www.oracle.com/technical-resources/articles/javase/locale.html#using).

The build is green (on my machine) and `upload-release.sh` also does not complain. I do not know what offer symptoms to look out for that might indicate that this change breaks anything.


tools/grammar/src/main/resources/log4j.xml
```diff
@@ -1,19 +1,14 @@
...
-		<level value="warn" />
-	</logger>
-	
-	<root>
-		<priority value="warn" />
-		<appender-ref ref="consoleAppender" />
-	</root>
-</log4j:configuration>
\ No newline at end of file
+<Configuration>
```
thobe, line 2: Should the xml name space be retained?

tools/grammar/src/test/resources/log4j.xml
```diff
@@ -1,19 +1,14 @@
...
-		<level value="warn" />
-	</logger>
-	
-	<root>
-		<priority value="warn" />
-		<appender-ref ref="consoleAppender" />
-	</root>
-</log4j:configuration>
\ No newline at end of file
+<Configuration>
```
thobe, line 2: I question whether this file is needed... I believe the `main/resources` are available to test executions as well, only that the `test/resources` can be used to override or provide additional resources.

Not that this is related to the change in this PR. I just happen to notice it since this PR made ma aware if the existence of this file.

thobe (approved):
My general sense is that if the build passes, then these changes are fine.

I'm thinking that it might be a good idea to get onto only a single version of JUnit, and not use both JUnit 4 for some things and JUnit 5 for some thing. But that can be a later change.

hvub:
@thobe thanks for the review. I have addressed your comment as follows:

 * I have revisited the junit4 topic. It is indeed not needed. Seems like a lost update on my part. I have removed it now. 
 * On `log4j.xml` namespace, I am not knowledgable on this. I have followed [these migration examples](https://logging.apache.org/log4j/2.x/manual/migration.html) and they do not keep the name space definition. 
 * I have removed `tools/grammar/src/test/resources/log4j.xml` and things still build. Always happy to something clean up.

I take, you still approve. Thanks.

thobe (approved):
I like it!

dwitry (approved):

PR #457, opened 2020-10-14
Categorize WITH clause

hvub:
This PR categorizes scenario on the `WITH` clause. It is not concerned with subclauses `ORDER BY`, `SKIP`, `LIMIT`. They will be categorized separately.

The PR renames quite a number of scenarios have been renamed in the process.

In total 7 scenario from `WithAcceptance` and 9 scenarios from `precategorized` have be categorized in to the following features:

| Feature | #scenarios |
|---|---|
| With1 - Forward single variable | 6 |
| With2 - Forward single expression | 3 |
| With3 - Forward multiple expressions | 1 |
| With4 - Variable aliasing | 4 |
| With5 - Implicit grouping with DISTINCT | 1 |
| With6 - Implicit grouping with aggregates | 4 |
| With7 - WITH on WITH | 2 |

This give a total of 21 scenarios in the `clause/with` category. 

On the side the PR renames `Match7 - Optional match scenarios` to `Match7 - Optional match` and  `Match8 - Match clause Interoperation with other clauses` to `Match8 - Match clause interoperation with other clauses`.

The category `clause/match` gains 4 additional scenarios. These scenarios contain a `WITH` clause, but actually test match behavior.

The category `precategorized` gain 2 scenarios.

Overall, the total gain of 11 scenarios results from materializing missing variations that became obvious in the categorization.


DvirDukhan:
@hvub 
Regarding the scenarios that test the interoperability of `WITH` with `ORDER BY`, `SKIP` and `LIMIT` - if you prefer to leave them where they are then we can merge.

hvub:
Yes, the idea is to categorize the subclause `ORDER BY`, `SKIP`, `LIMIT` separately. I will merge.

PR #460, opened 2020-11-05
Remove invalid construct YIELD *

jeffreylovitz:
This PR modifies the grammar rule that allows the `YIELD *` construct, which is not recognized by Neo4j as valid syntax:
```
neo4j@neo4j> CALL db.labels() YIELD *;
Invalid input '*': expected whitespace, comment, aliased procedure result field or simple procedure result field (line 1, column 24 (offset: 23))
"CALL db.labels() YIELD *;"
```

### DIFF
--- a/grammar/cypher.xml
+++ b/grammar/cypher.xml
@@ -274,13 +274,10 @@
 
   <production name="YieldItems" rr:inline="true">
     <seq>
-      <alt>
-        *
-        <seq>
-          <non-terminal ref="YieldItem"/>
-          <repeat>&WS; , &WS;<non-terminal ref="YieldItem"/></repeat>
-        </seq>
-      </alt>
+      <seq>
+        <non-terminal ref="YieldItem"/>
+        <repeat>&WS; , &WS;<non-terminal ref="YieldItem"/></repeat>
+      </seq>
       <opt>&WS; <non-terminal ref="Where"/></opt>
     </seq>
   </production>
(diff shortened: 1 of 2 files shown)


thobe:
Thank you for this! According to [the specification](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-06-24-call-procedures.adoc#8-appendix-procedure-call-patterns), `YIELD *` should only be allowed in standalone calls. Although not even that is supported by Neo4j, which suggests that it might not be a very useful feature.

jeffreylovitz:
Oh, interesting! I just came across this and assumed it was an error; feel free to close the PR if you prefer to leave the specification unmodified.

thobe:
No, I think you highlight a real inconsistency. But I note that there might be more that needs fixing than what you changed.

Given that `YIELD *` for standalone calls is not implemented by Neo4j, and not asked for, I'm wondering if it is really needed. So possibly we should change the specification.

I'll leave this open until there has been time to have a discussion around this, and then probably either merge it or supersede it with another change that fixes the related issues as well.

A simple one that I note is that `YieldItems` allows for a sequence of a sequence of items followed by an optional where clause - that is one level of sequence nesting too many.

hvub:
To progress this I have created #497. If people are happy with #497 and is considered as a satisfactory solution, this PR here can probably be closed, eventually.

hvub:
I have merged #497 and subsequently will close this PR.

PR #463, opened 2020-11-11
Categorize with where

hvub:
This PR categorizes scenario testing the `WHERE` sub-clause of the `WITH` clause. The categorize are as follows:

| Feature | Number of scenarios |
| --- | --- |
| WithWhere1 - Filter single variable | 4	|
| WithWhere2 - Filter multiple variables | 2 |
| WithWhere3 - Equi-Joins on variables | 3 |
| WithWhere4 - Non-Equi-Joins on variables | 2 |
| WithWhere5 - Filter on predicate resulting in null | 4 |
| WithWhere6 - Filter on aggregates | 1 |
| WithWhere7 - Variable visibility under aliasing | 3 |

The scenarios in WithWhere1 and WithWhere6 are scenarios taken from the old `WithAccceptance.feature`. Some of them renamed.

The scenarios in WithWhere2 – 5 are copied from ReturnWhere2 – 5, respectively, and modified by inserting a `WITH` clause before `WHERE`.

The scenarios in WithWhere7 are inspired by WithWhere1.[2]. They are variations of one-another, that test which variables names are visible in `WHERE`.

The PR deliberately does not consider negative scenarios to avoid merge conflicts with #458. If there are uncategorized negative scenarios left after merging #458 that belong to WithWhere, this PR, if still open, can be updated or a separate PR can be created.


DvirDukhan (commented):
Looks good, small fix

PR #347, opened 2019-02-06
Add CIP on SET and REMOVE

sherfert:
https://github.com/sherfert/openCypher/blob/8051aa6b23d8e4b3e24d190d7b08ef796c52f090/cip/1.accepted/CIP2019-02-06-SET.adoc

PR #464, opened 2020-11-16
Change Type system CIP w.r.t. Maps with null values.

sherfert:


cip/1.accepted/CIP2015-09-16-public-type-system-type-annotation.adoc
```diff
@@ -128,17 +128,24 @@ a nullable variant:
...
 * `MAP`
-** Maps from string keys to values of any type, i.e. `{ name: 'Svensson' }`. The type of map values is `ANY?` since it is unknown to the type system if a given map contains a certain key or not. Note also that maps do not distinguish between missing keys and keys that map to a `null` value. `MAP` is a parent to all property containers, like `NODE` and `RELATIONSHIP`.
+** Maps from string keys to values of any type, i.e. `{ name: 'Svensson' }`.
+   The type of map values is `ANY?` since it is unknown to the type system if a given map contains a certain key or not.
+   Note also that maps distinguish between missing keys and keys that map to a `null` value.
+   As an example, that means that `{name: null}` is not equal to `{}`, the empty map.
+   A use case where this distinction is particularly important is when setting multiple properties on a node.
+   `SET n += {name: null, age: 20}` will remove any existing `name` properties while `SET n += {age: 20}` would leave `name` properties untouched.
```
sherfert, line 136: I think it is OK the reference `SET` as a use case here. Even though the [SET CIP](https://github.com/opencypher/openCypher/pull/347) has not been merged so far (and also does not specify this behavior), we have a [TCK scenario](https://github.com/sherfert/openCypher/blob/maps/tck/features/clauses/set/Set5.feature#L83) which is consistent with this description.

cip/1.accepted/CIP2015-09-16-public-type-system-type-annotation.adoc
```diff
@@ -128,17 +128,24 @@ a nullable variant:
...
 a nullable variant:
 
 * `NODE`
-** A `NODE` is a node in the property graph model with properties, labels, and relationships. A node is an entity and a property container in the model and therefore also a `MAP`.
+** A `NODE` is a node in the property graph model with properties, labels, and relationships.
+   A node is an entity and a property container.
 * `RELATIONSHIP`
-** A `RELATIONSHIP` is a relationship in the property graph model with properties, relationship type, a start node, and an end node. A relationship is an entity and a property container in the model and therefore also a `MAP`.
+** A `RELATIONSHIP` is a relationship in the property graph model with properties, relationship type, a start node, and an end node.
+   A relationship is an entity and a property container.
```
Mats-SX, line 148: I love that you just took that out. I am in favour. I do not know if this particular change has any repercussions, but I don't think so. It would be up to the implementation(s) to find out where and when it would not be fine to treat elements/entities as maps.

Mats-SX (approved):
Looks very neat!

hvub (approved):
I have created a card for myself to check the distinction between missing keys and keys that map to a `null` value is/can be codified in the TCK.

Mats-SX:
@hvub I'm sure you have some ideas, but I can share these simple tests just in case you need inspiration:

```
{} = {k: null}
{k: 1} = {k: 1, l: null}
{k: null} = {k: null, l: null}

CREATE (n) SET n += {} // nop
CREATE (n {k: 1}) SET n += {k: null}
```

PR #470, opened 2020-12-01
Scenarios for maps with null values

hvub:
This PR follows up #464 with changing and augmenting TCK scenarios to reflect the behavior now clarified in the [CIP](https://github.com/sherfert/openCypher/blob/71f865baba103b54ad7bda72c360917c9c4ccc17/cip/1.accepted/CIP2015-09-16-public-type-system-type-annotation.adoc).

### DIFF
--- a/tck/features/clauses/set/Set4.feature
+++ b/tck/features/clauses/set/Set4.feature
@@ -67,7 +67,7 @@ Feature: Set4 - Set all properties with a map
       | +properties | 2 |
       | -properties | 2 |
 
-  Scenario: [2] Null values in a property map are removed with SET
+  Scenario: [3] Null values in a property map are removed with SET
     Given an empty graph
     And having executed:
       """
@@ -86,7 +86,7 @@ Feature: Set4 - Set all properties with a map
       | +properties | 2 |
       | -properties | 2 |
 
-  Scenario: [3] All properties are removed if node is set to empty property map
+  Scenario: [4] All properties are removed if node is set to empty property map
     Given an empty graph
     And having executed:
       """
--- a/tck/features/clauses/set/Set4.feature
+++ b/tck/features/clauses/set/Set4.feature
@@ -28,7 +28,7 @@
 
 #encoding: utf-8
 
-Feature: Set4 - Set All Properties with a Map
+Feature: Set4 - Set all properties with a map
 
   Scenario: [1] Set multiple properties with a property map
     Given an empty graph
@@ -48,7 +48,7 @@ Feature: Set4 - Set All Properties with a Map
     And the side effects should be:
       | +properties | 3 |
 
-  Scenario: [2] Non-existent values in a property map are removed with SET =
+  Scenario: [2] Non-existent values in a property map are removed with SET
     Given an empty graph
     And having executed:
       """
@@ -61,7 +61,26 @@ Feature: Set4 - Set All Properties with a Map
       RETURN n
       """
     Then the result should be, in any order:
-      | n                           |
+      | n                          |
+      | (:X {name: 'B', baz: 'C'}) |
+    And the side effects should be:
+      | +properties | 2 |
+      | -properties | 2 |
+
+  Scenario: [2] Null values in a property map are removed with SET
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (:X {name: 'A', name2: 'B'})
+      """
+    When executing query:
+      """
+      MATCH (n:X {name: 'A'})
+      SET n = {name: 'B', name2: null, baz: 'C'}
+      RETURN n
+      """
+    Then the result should be, in any order:
+      | n                          |
       | (:X {name: 'B', baz: 'C'}) |
     And the side effects should be:
       | +properties | 2 |
@@ -85,7 +104,7 @@ Feature: Set4 - Set All Properties with a Map
     And the side effects should be:
       | -properties | 2 |
 
-  Scenario: [4] Ignore null when setting properties using an overriding map
+  Scenario: [5] Ignore null when setting properties using an overriding map
     Given an empty graph
     When executing query:
       """
--- a/tck/features/expressions/comparison/Comparison1.feature
+++ b/tck/features/expressions/comparison/Comparison1.feature
@@ -143,7 +143,37 @@ Feature: Comparison1 - Equality
       | [[1], [2]]    | [[1], [null]] | null   |
       | [[1], [2, 3]] | [[1], [null]] | false  |
 
-  Scenario Outline: [7] Equality and inequality of NaN
+  Scenario Outline: [7] Comparing maps to maps
+    Given an empty graph
+    When executing query:
+      """
+      RETURN <lhs> = <rhs> AS result
+      """
+    Then the result should be, in any order:
+      | result   |
+      | <result> |
+    And no side effects
+
+    Examples:
+      | lhs             | rhs                | result |
+      | {}              | {}                 | true   |
+      | {k: true}       | {k: true}          | true   |
+      | {k: 1}          | {k: 1}             | true   |
+      | {k: 1.0}        | {k: 1.0}           | true   |
+      | {k: 'abc'}      | {k: 'abc'}         | true   |
+      | {k: 'a', l: 2}  | {k: 'a', l: 2}     | true   |
+      | {}              | {k: null}          | false  |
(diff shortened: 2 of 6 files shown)


hvub:
I will fix the red build soon. No rush.

hvub:
Ready to be reviewed.

sherfert (approved):
The changes look good to me.

Before merging I would run the new scenarios with the latest neo4j.
This will make sure that we didn't make a mistake by accident
or it will uncover any bugs in the implementation earlier.

tck/features/expressions/comparison/Comparison1.feature
```diff
@@ -143,7 +143,37 @@ Feature: Comparison1 - Equality
...
+      | {}              | {k: null}          | false  |
+      | {k: null}       | {}                 | false  |
+      | {k: 1}          | {k: 1, l: null}    | false  |
+      | {k: null, l: 1} | {l: 1}             | false  |
+      | {k: null}       | {k: null, l: null} | false  |
+      | {k: null}       | {k: null}          | null   |
+      | {k: 1}          | {k: null}          | null   |
+      | {k: 1, l: null} | {k: null, l: null} | null   |
+      | {k: 1, l: null} | {k: null, l: 1}    | null   |
+      | {k: 1, l: null} | {k: 1, l: 1}       | null   |
```
Mats-SX, line 174: 👍

tck/features/expressions/map/Map3.feature
```diff
@@ -65,3 +65,25 @@ Feature: Map3 - Keys function
...
+      """
+      RETURN keys(<map>) AS keys
+      """
+    Then the result should be (ignoring element order for lists):
+      | keys     |
+      | <result> |
+    And no side effects
+
+    Examples:
+      | keys                  | result          |
```
Mats-SX, line 81:
```suggestion
      | map                   | result          |
```

tck/features/expressions/map/Map3.feature
```diff
@@ -65,3 +65,25 @@ Feature: Map3 - Keys function
...
+      | {}                    | []              |
+      | {k: 1}                | ['k']           |
+      | {k: null}             | ['k']           |
+      | {null: null}          | ['null']        |
```
Mats-SX, line 85: 🙈

Mats-SX: I see the point of adding this here, but technically the key `null` is no different from the key `k`. It is amusing, but it is potentially confusing, by making an unaware reader think that Cypher supports some kind of odd concept of `null` keys... which we don't. Maybe we should not keep this?

tck/features/expressions/map/Map4.feature
```diff
@@ -35,18 +35,28 @@ Feature: Map4 - Field existence check
...
-      | {name: 'Mats', name2: 'Pontus'} | true   |
-      | {name: null}                    | false  |
-      | {notName: 0, notName2: null}    | false  |
+      | map                                 | key     | result |
+      | {name: 'Mats', name2: 'Pontus'}     | name    | true   |
+      | {name: 'Mats', name2: 'Pontus'}     | name2   | true   |
+      | {name: null}                        | name    | true   |
+      | {name: null, name2: 'Pontus'}       | name    | true   |
+      | {name: null, name2: null}           | name    | true   |
+      | {name: null, name2: null}           | name2   | true   |
```
Mats-SX, line 52: what is the difference between these two cases? Isn't one redundant?

tck/features/expressions/map/Map4.feature
```diff
@@ -35,18 +35,28 @@ Feature: Map4 - Field existence check
...
+      | {name: 'Pontus', name2: null}       | name2   | true   |
```
Mats-SX, line 53: this looks identical to line 50; one null value and one non-null, and then we reach for the key with null value.

tck/features/expressions/map/Map4.feature
```diff
@@ -35,18 +35,28 @@ Feature: Map4 - Field existence check
...
+      | {name: 'Pontus', notName2: null}    | name    | true   |
+      | {notName: 'Pontus', notName2: null} | notName | true   |
```
Mats-SX, line 55: same as line 54

tck/features/expressions/map/Map4.feature
```diff
@@ -35,18 +35,28 @@ Feature: Map4 - Field existence check
...
+      | {notName: null, notName2: null}     | notName | true   |
```
Mats-SX, line 56: same as line 51

tck/features/expressions/map/Map4.feature
```diff
@@ -35,18 +35,28 @@ Feature: Map4 - Field existence check
...
+      | {notName: null, notName2: null}     | name    | false  |
+      | {notName: 0, notName2: null}        | name    | false  |
+      | {notName: 0}                        | name    | false  |
```
Mats-SX, line 59: I think we should also make sure to have a case with an empty map

Mats-SX (changes requested):
I like the comprehensiveness of the PR. I found two bugs and some duplicated test cases. I also have some reservations around using `{null: null}` as an example case; I think it isn't helpful, and I hope we can even make it invalid.

tck/features/clauses/set/Set4.feature
```diff
@@ -61,7 +61,26 @@ Feature: Set4 - Set All Properties with a Map
...
       """
     Then the result should be, in any order:
-      | n                           |
+      | n                          |
+      | (:X {name: 'B', baz: 'C'}) |
+    And the side effects should be:
+      | +properties | 2 |
+      | -properties | 2 |
+
+  Scenario: [2] Null values in a property map are removed with SET
```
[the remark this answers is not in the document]
hvub: Yeah, could be turned into an outline, the name would become less clear. I rather keep them separated and fix the numbering. Thanks for spotting.

tck/features/expressions/map/Map4.feature
hvub: Well the second column is. Side-by-side, the two examples show, that it does not matter which field you test existence on or where this field in the order given by the literal.

tck/features/expressions/map/Map4.feature
hvub: Again, the order of field does not matter. If you say, these scenarios are identical, you implicitly assume an implementation follows the rule that with field order does not matter. Go tests do not make such assumptions.

tck/features/expressions/map/Map4.feature
hvub: agree

tck/features/expressions/map/Map4.feature
hvub: agree

tck/features/expressions/map/Map4.feature
hvub: absolutely.

tck/features/expressions/map/Map3.feature
hvub: Well, where are different opinions on this topic and currently `null` is allowed, so it make sense to test that it does not confuse the keys function. It is the corner cases that are interesting for testing and blow up in the wild at some point if they have not been tested.

But since the topic of reserved keyword is still controversial, I do not mind removing the example.

hvub:
@sherfert and @Mats-SX, thank you for your timely reviews.

@Mats-SX, I hope I have addressed all you comments sufficiently, where I agree with them. You may want to give it a second look and react to it.

tck/features/expressions/map/Map4.feature
Mats-SX: I would've thought it was another test's responsibility to check that field order in maps have no meaning, but I'm happy to keep the scenario given your reasoning.

tck/features/expressions/map/Map3.feature
Mats-SX: I mean, the only reason you wrote this is to illustrate that `null` is treated in no special way when written in a map key position; in effect it is unrelated to what this scenario is testing for. I'm not necessarily against a `MapLiteral` feature scenario including both this literal as well as `{true: false, false: true, `1`: 1}` kind of maps, but then I would say it should be clear in the scenario description that it deals with the topic of map keys.

Mats-SX (approved):

Mats-SX:
@hvub Thanks for the changes -- LGTM 👍

PR #471, opened 2020-12-01
Remove VOID from procedure signatures

hvub:
The PR 

 * changes procedure signatures in TCK scenarios that have return type `VOID` to return type `()` and 
 * removes duplicate scenarios that result from this change.

Return type `VOID` to return type `()` do not have a semantics difference in openCypher at the moment. The [Call CIP](https://github.com/openCypher/openCypher/blob/master/cip/1.accepted/CIP2015-06-24-call-procedures.adoc) does not allow `VOID` as output signature. Hence, there is no point in keeping such scenarios.

### DIFF
--- a/cip/1.accepted/CIP2015-09-16-public-type-system-type-annotation.adoc
+++ b/cip/1.accepted/CIP2015-09-16-public-type-system-type-annotation.adoc
@@ -50,23 +50,23 @@ The details of internal types are out of scope for the purposes of this document
 
 === Types and Type Literal Syntax
 
-The section below presents presents the new Public Type System (PTS) proposed for Cypher.
+The section below presents the new Public Type System (PTS) proposed for Cypher.
 
 ==== Material and nullable types
 
 The type system provides a way to track nullability, i.e. a type may express if a given expression may be `null` or not.
-Tracking (non-)nullability helps an implementation to detect and warn about possible user errors during compile time and  it also enables optimizations that rely on values not being `null`. As an example, an implementation may infer that `MATCH` will never introduce a nullable variable, but `OPTIONAL MATCH` will only introduce nullable variables.
-
-Material types are all types that do not permit `null` as a valid result of evaluating the underlying expression.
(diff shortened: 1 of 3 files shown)


thobe (changes requested):
https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc talks about `VOID` as the type of `null` - that is made a lot clearer by this change, but we might still consider a different name for the type of null at some point.

However https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-09-16-public-type-system-type-annotation.adoc#218-void-type also talks about the `VOID` type as a type that only has `null` as value, but explicitly talks about this being used for procedure output signatures - this should change.

hvub:
@thobe I have made the minimal changes to the CIPs to remove `VOID`, I hope. (... and fixed two typos on the way)

If you look a little closer, you will see that I also changed the nullable vs. mandatory logic, i.e. `T?` vs `T!`, first. The CIP is inconsistent with what we came up for CREATE INDEX. However, the CIP is consistent with what [APOC does around procedure signatures](https://neo4j.com/developer/neo4j-apoc/#_call_apoc_helpkeyword). A quick investigation with

```
CALL apoc.help('') YIELD name, signature
WITH name, (signature CONTAINS "?") as opt
WHERE opt
RETURN COUNT(DISTINCT name) as c
UNION ALL
CALL apoc.help('') YIELD name, signature
RETURN COUNT(DISTINCT name) as c
```

revealed, though, that all signatures contain at least one nullable type, which suggests that nullable as default would be the better choice here too.

Anyway, I decide that this is a topic in and of itself and reverted these changes here in this PR.

hvub:
@thobe did you had a change to look at the changes I made up on your review? What are the next steps?

thobe (approved):
@hvub Sorry, I missed the notification of you having made these changes. This looks good now.

PR #481, opened 2021-01-20
Facilitate named examples in scenario outlines

hvub:
(This PR is based on PR #478)

This PR extends the TCK API and related tools to support naming of examples in scenario outlines.

The PR modifies the `TypeConversion1.[5]` scenario outline to illustrate the facility:

```Gherkin
  @NegativeTest
  Scenario Outline: [5] `toBoolean()` on invalid types #Example: <exampleName>
    Given any graph
    When executing query:
      """
      WITH [true, <invalid>] AS list
      RETURN toBoolean(list[1]) AS b
      """
    Then a TypeError should be raised at runtime: InvalidArgumentValue

    Examples:
      | invalid | exampleName |
      | []      | list        |
      | {}      | map         |
      | 1       | integer     |
      | 1.0     | float       |
```

The outline parameter `<exampleName>` does not contribute to the scenario other than modifying the name of the outline. Note that the parameter in the outline name appears after ` #Example: `. With the change made by this PR, the TCK API detects this pattern.

The `Scenario` has a new field `exampleName: Optional[String]`. This field is set to `None` if the pattern is not used. If the pattern is used, `exampleName` is set to `Some(x)`, where `x` is whatever string comes after the marker ` #Example: `.

For instance, for first example the `TypeConversion1.[5]` scenario outline `exampleName` will be set to the string `list`; for the second example to `map`; and so on. The `name` field in all scenarios resulting from this outline will be set to `` `toBoolean()` on invalid types``, i.e. everything between the number (`[5] `) and the ` #Example: ` marker.

Note that independently of this PR and the use of this pattern, `Scenario` objects have a field `exampleIndex: Option[Int]` that contains the running number (starting at `0`) for all scenarios resulting from the examples of the same outline in the order of the example listing in the feature file.

Since scenario resulting outline have an example index independently from the example name, the example name does not have to be unique, i.e. the same name can be used for multiple examples in the same outline. For instance, `TypeConversion1.[5]` could be modified as follow and would still be valid:

```Gherkin
  @NegativeTest
  Scenario Outline: [5] `toBoolean()` on invalid types #Example: <exampleName>
    Given any graph
    When executing query:
      """
      WITH [true, <invalid>] AS list
      RETURN toBoolean(list[1]) AS b
      """
    Then a TypeError should be raised at runtime: InvalidArgumentValue

    Examples:
      | invalid | exampleName |
      | []      | list        |
      | [1]     | list        |
      | {}      | map         |
      | 1       | integer     |
      | 1.0     | float       |
```

Note that the first and the second example both have the name `list`.

The name of the outline parameter used to provide the example name does not matter, other than is it have to be different from the name of the other outline parameters use in a given outline. It is very sensible to call it `<exampleName>`, but this is not required. For instance `<abc>` would work equally well with regards to the TCK API. (In fact, the TCK API does not even see the parameter name, since outline examples are already instantiated by the Cucumber's Gherkin parser.)

The ScenarioFormatChecker checks that [every scenario with an example name also has an example index](https://github.com/opencypher/openCypher/compare/master...hvub:facilitate-named-examples-in-scenario-outlines?expand=1#diff-cccf4151c32c8d59e8d1a5742857c1f0d3ee94a11e5d363da51badd5f4bc53cdR58-R64). In other words, the marker ` #Example: ` is reserved for providing example names in scenario outlines and prohibited in regular non-outline scenario names.

### DIFF
--- a/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
+++ b/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
@@ -31,7 +31,7 @@ import java.net.URI
 import java.nio.charset.StandardCharsets
 import java.nio.file._
 import java.util
-
+import io.cucumber.core.gherkin
 import io.cucumber.core.gherkin.DataTableArgument
 import io.cucumber.core.gherkin.DocStringArgument
 import io.cucumber.core.gherkin.vintage.GherkinVintageFeatureParser
@@ -50,6 +50,7 @@ import scala.collection.JavaConverters._
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
+import scala.util.matching.Regex
 
 object CypherTCK {
 
@@ -120,9 +121,26 @@ object CypherTCK {
     parseFeature(featureFile, featureString, categories)
   }
 
-  private val pickleNamePattern1 = "^(((?! #Example: ).)+)(?: #Example: (.*)?)?$".r
-
   def parseFeature(featureFile: Path, featureString: String, categories: Seq[String]): Feature = {
+
+    case class NameExtractedPickle(pickle: gherkin.Pickle, nameAndNumber: String, exampleName: Option[String])
+
+    object parsePickleName extends (gherkin.Pickle => NameExtractedPickle) {
+      private val exampleNamePattern = "^(((?! #Example: ).)+)(?: #Example: (.*)?)?$".r
+
+      def apply(pickle: gherkin.Pickle): NameExtractedPickle = pickle.getName match {
+        case exampleNamePattern(nameAndNumber, _, null) => NameExtractedPickle(pickle, nameAndNumber, None)
+        case exampleNamePattern(nameAndNumber, _, exampleName) => NameExtractedPickle(pickle, nameAndNumber, Some(exampleName))
+        case _ => NameExtractedPickle(pickle, pickle.getName, None)
+      }
+    }
+
+    case class PickleGroupingKey(keyword: String, pickleName: String)
+
+    object formPickleGroupingKey extends (NameExtractedPickle => PickleGroupingKey) {
+      def apply(pickle: NameExtractedPickle): PickleGroupingKey = PickleGroupingKey(pickle.pickle.getKeyword, pickle.nameAndNumber)
+    }
+
     Try(parser.parse(featureFile.toUri, featureString, null)) match {
       case Success(featureOption) =>
         if(featureOption.isPresent) {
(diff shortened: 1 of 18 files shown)


tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
```diff
@@ -129,18 +131,25 @@ object CypherTCK {
           // filters out scenarios with TCKTags.IGNORE
           val included = pickles.filterNot(tagNames(_) contains TCKTags.IGNORE)
 
-          val includedGroupedByName = included.groupBy(_.getName)
-          val includedGroupedAndSorted = includedGroupedByName.mapValues(_.sortBy(_.getLocation.getLine))
+          val includedWithExtractedName = included.map(p => {
+            p.getName match {
+              case pickleNamePattern1(pickleName, _, null) => (p, pickleName, None)
+              case pickleNamePattern1(pickleName, _, exampleName) => (p, pickleName, Some(exampleName))
```
mnd999, line 137: Is it worth creating a case class here, ._1, ._2, ._3 are not very descriptive?

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
```diff
@@ -129,18 +131,25 @@ object CypherTCK {
...
-              if(pickles.size > 1)
+            case (("Scenario Outline", _), pickles) =>
                 pickles.zipWithIndex.map {
-                  case (pickle, exampleIndex) => toScenario(categories, featureName, pickle, Some(exampleIndex), featureFile)
+                  case (pickle, exampleIndex) => toScenario(categories, featureName, pickle._2, Some(exampleIndex), pickle._3, pickle._1, featureFile)
                 }
-              else
-                pickles.map(pickle => toScenario(categories, featureName, pickle, None, featureFile))
+            case (("Scenario", _), pickles) =>
+                pickles.map(pickle => toScenario(categories, featureName, pickle._2, None, pickle._3, pickle._1, featureFile))
```
mnd999, line 152: pickle._3 (the outline example name - I think) is always None here isn't it. Perhaps replace with None.

tools/tck-inspection/src/test/scala/org/opencypher/tools/tck/inspection/diff/GroupDiffTest.scala
```diff
@@ -212,9 +212,9 @@ class GroupDiffTest extends AnyFunSuite with Matchers {
   test("Diff a group with one scenario changed in categories, tags, and content of steps") {
     val stepsA1 = List[Step](Dummy(dummyPickleStep), Measure(dummyPickleStep))
     val stepsA2 = List[Step](Measure(dummyPickleStep), Dummy(dummyPickleStep))
-    val scrA1 = Scenario(List[String](), "ftr1", Some(1), "scrA", None, Set[String]("T"), stepsA1, dummyPickle, dummyPath("ftr1.feature"))
-    val scrA2 = Scenario(List[String](), "ftr2", Some(1), "scrA", None, Set[String]("X"), stepsA2, dummyPickle, dummyPath("ftr2.feature"))
-    val scrB = Scenario(List[String](), "ftr1", Some(2), "scrB", None, Set[String](), List[Step](), dummyPickle, dummyPath("ftr1.feature"))
+    val scrA1 = Scenario(List[String](), "ftr1", Some(1), "scrA", None, None, Set[String]("T"), stepsA1, dummyPickle, dummyPath("ftr1.feature"))
+    val scrA2 = Scenario(List[String](), "ftr2", Some(1), "scrA", None, None, Set[String]("X"), stepsA2, dummyPickle, dummyPath("ftr2.feature"))
+    val scrB = Scenario(List[String](), "ftr1", Some(2), "scrB", None, None, Set[String](), List[Step](), dummyPickle, dummyPath("ftr1.feature"))
     val result = GroupDiff(Seq(scrA1, scrB), Seq(scrA2, scrB))
```
mnd999, line 219: Do we need a test here showing diff counts for example name only changing? Or are we happy this is covered by ""Diff equal scenarios differing in example name only" below?

mnd999 (approved):
Just a few improvement suggestions, otherwise looks great. Nothing blocking though as far as I can see, so will approve.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
hvub: Could. Yes. But I like to pass it through. It does not bother scenario execution at all. However, the ScenarioFormatChecker see it and [can test for it](https://github.com/opencypher/openCypher/pull/481/files#diff-cccf4151c32c8d59e8d1a5742857c1f0d3ee94a11e5d363da51badd5f4bc53cdR55-R62). If we would set the example to `None` here, a non-outline scenario illegally using the ` #Example: ` marker would remain undiscovered. The API would simple gloss over it. The ScenarioFormatChecker would not see. 

Now, one could discuss if the API should not do all the integrity checks in the first place, why have a separate ScenarioFormatChecker anyway? That is decision I was not involved in, but I reckon a good reason to have the separation is to keep the API rather lean. The API is ran much more frequently – when ever scenarios are parsed in to run compatibility tests. The ScenarioFormatChecker only needs to be ran when scenarios have been changed.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
```diff
@@ -151,16 +160,16 @@ object CypherTCK {
     }
   }
 
-  private val pickleNamePattern = "^\\Q[\\E([0-9]+)\\Q]\\E (.+)$".r
+  private val pickleNamePattern2 = "^\\Q[\\E([0-9]+)\\Q]\\E (.+)$".r
```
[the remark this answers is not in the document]
hvub: I will address this.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
hvub: I will address this.

tools/tck-inspection/src/test/scala/org/opencypher/tools/tck/inspection/diff/GroupDiffTest.scala
hvub: I will add one.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
mnd999: Sure, sounds reasonable. I hadn't considered that the example name would make it through the parsing to this point if the test is a Scenario instead of a Scenario Outline.

PR #489, opened 2021-03-24
add scenarios to call5

DvirDukhan:


tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,102 @@
...
+  Scenario: [6] Yield wrong type - integer as string
+    Given an empty graph
+    And there exists a procedure test.my.proc(in :: INTEGER?) :: (out :: STRING?):
+      | in   | out   |
+      | null | 1 |
+    When executing query:
+      """
+      CALL test.my.proc(null)
+      """
+    Then a SyntaxError should be raised at compile time: InvalidArgumentType
```
swilly22, line 118: I'm not sure I follow scenarios 6 and 7

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,102 @@
...
+    And no side effects
+
+  Scenario: [3] Explicit procedure result projection with RETURN *
+    Given an empty graph
+    And there exists a procedure test.my.proc(in :: INTEGER?) :: (out :: STRING?):
+      | in   | out   |
+      | null | 'nix' |
+    When executing query:
+      """
+      CALL test.my.proc(null) YIELD out
```
swilly22, line 70: Please add a scenario in which the procedure have multiple outputs
validate that output reordering via YIELD, e.g. `CALL P() YIELD A, B` and `CALL P() YIELD B, A`

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,102 @@
 #encoding: utf-8
 
 Feature: Call5 - Results projection
+
```
swilly22, line 32: What about specifying an expression as part of YIELD? I guess that should trigger and error,
`YIELD A+1`

Same idea for yielding none existing outputs

swilly22 (changes requested):

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,102 @@
...
+  Scenario: [7] Yield wrong type - string as integer
+    Given an empty graph
+    And there exists a procedure test.my.proc(in :: INTEGER?) :: (out :: INTEGER?):
+      | in   | out   |
+      | null | 'nix' |
+    When executing query:
+      """
+      CALL test.my.proc(null)
+      """
+    Then a SyntaxError should be raised at compile time: InvalidArgumentType
```
thobe, line 130: I am not sure that procedures that violate their own signature is within the scope of the language specification.

It seems to me like these test cases might be triggering an exception in the testing framework rather than in an implementation.

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,109 @@
...
+      | null | 1 | 2 |
+    When executing query:
+      """
+      CALL test.my.proc(null) YIELD A, B
+      WITH out as a RETURN a
+      """
+    Then the result should be, in order:
+      | A | B |
+      | 1 | 2 |
+    And no side effects
```
hvub, line 121: This scenario is wrong. The one below too. With `RETURN a` the result table will only have `a` as its single column.

With regards to my comment above about the order of yield items is irrelevant, I would change the scenario as follows:

```suggestion
  Scenario Outline: [6] The order of yield items is irrelevant
    Given an empty graph
    And there exists a procedure test.my.proc(in :: INTEGER?) :: (A :: INTEGER?, B :: INTEGER?) :
      | in   | a | b |
      | null | 1 | 2 |
    When executing query:
      """
      CALL test.my.proc(null) YIELD <yield>
      RETURN a, b
      """
    Then the result should be, in order:
      | a | b |
      | 1 | 2 |
    And no side effects
    
    Examples:
      | yield |
      | a, b  |
      | b, a  |
```

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,109 @@
...
+      | null | 1 | 2 |
+    When executing query:
+      """
+      CALL test.my.proc(null) YIELD B, A
+      WITH out as a RETURN a
+      """
+    Then the result should be, in order:
+      | B | A |
+      | 2 | 1 |
+    And no side effects
```
hvub, line 136: I would remove this scenario. It is wrong and does not add, as explained in the two comments above.
```suggestion
```

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,109 @@
...
+      | null | 'nix' |
+    When executing query:
+      """
+      CALL test.my.proc(null) YIELD out
+      WITH out as a RETURN a
+      """
+    Then the result should be, in order:
+      | a     |
+      | 'nix' |
+    And no side effects
```
hvub, line 106: I do not understand what these two scenario are aiming to test. If it is around the interplay of `CALL` with a trailing `WITH` clause it should probably be placed in `Call6 - Call clause interoperation with other clauses`.

hvub (changes requested):
Thanks for the PR. I made a couple of comments, including replies to @swilly22 comments. Looks to me this is a good starting point, but need more work. I hope the suggestions in the comments are helpful. Happy to discuss more in the next meeting, if more clarification or so is needed.

Something not mentioned in the comments below: CI the fails on the [ScenarioFormatChecker](https://github.com/opencypher/openCypher/blob/master/tools/tck-integrity-tests/src/test/scala/org/opencypher/tools/tck/TckScenarioFormatCheck.scala), which can be ran locally. The CI output should only Cypher style violation. It needs a more detailed look what precisely these style violations are. Anyway, it something that should get fixed before the PR get merged, but is can happen ones the set of scenarios is fix. I can also help with it at that point.

Other aspects I learned:

 - If the procedure signature has return type `VOID`, the `YIELD` subclause is prohibit.
 - If the procedure signature has a return type other than `VOID`, the `YIELD` subclause is optional. Without `YIELD` you still get the cardinality effect but no new columns. (I doubt that that this to be tested in Call5, but we should make sure )

In general, we have a [CIP](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-06-24-call-procedures.adoc#223-in-statement-calls) and we should make sure to align with it as well as test as many aspects of the CIP as reasonable possible.

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,153 @@
...
+      """
+      CALL test.my.proc(null) YIELD a as b, b as a
+      RETURN b, a
+      """
+    Then the result should be, in order:
+      | b | a |
+      | 1 | 2 |
+    And no side effects
+
+  Scenario Outline: [9] Fail on renaming to an already bound variable name
```
hvub, line 158: This is good. There is also the case, when the variable has been bound before the CALL, like

- `MATCH (a)
CALL test.my.proc(null) YIELD a AS c, b AS a
RETURN a`
- `WITH 1 AS a
CALL test.my.proc(null) YIELD a AS c, b AS a
RETURN a`

The same error can be provoked without renaming:

- `MATCH (a) 
CALL test.my.proc(null) YIELD a, b 
RETURN a`
- `WITH 1 AS a
CALL test.my.proc(null) YIELD a, b
RETURN a`

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,153 @@
...
+      | null | 1 | 2 |
+    When executing query:
+      """
+      CALL test.my.proc(null) YIELD a as b, b as a
+      RETURN b, a
+      """
+    Then the result should be, in order:
+      | b | a |
+      | 1 | 2 |
+    And no side effects
```
hvub, line 156: The renaming scenario could be integrated into a scenario outline, I think:

```suggestion
  Scenario Outline: [6] Rename outputs to unbound variable names
    Given an empty graph
    And there exists a procedure test.my.proc(in :: INTEGER?) :: (A :: INTEGER?, B :: INTEGER?) :
      | in   | a | b |
      | null | 1 | 2 |
    When executing query:
      """
      CALL test.my.proc(null) YIELD <rename>
      RETURN <a>, <b>
      """
    Then the result should be, in order:
      | <a> | <b> |
      | 1   | 2   |
    And no side effects
    
    Examples:
      | <rename> | <a> | <b> |
      | a AS c, b AS d | c   | d   |
      | a AS b, b AS d | b   | d   |
      | a AS c, b AS a | c   | a   |
      | a AS b, b AS a | b   | a   |
      | a AS c, b AS b | c   | b   |
      | a AS c, b      | c   | b   |
      | a AS a, b AS d | a   | d   |
      | a, b AS d      | a   | d   |
      | a AS a, b AS b | a   | b   |
      | a AS a, b      | a   | b   |
      | a, b AS b      | a   | b   |
    
```

As illustrated, this makes iterating through a larger number of case very simple.

Similarly a the corresponding failing scenarios can be arrange in a single scenario outline.

hvub, line 156: The renaming scenario could be integrated into a scenario outline, I think:

```suggestion
  Scenario Outline: [6] Rename outputs to unbound variable names
    Given an empty graph
    And there exists a procedure test.my.proc(in :: INTEGER?) :: (A :: INTEGER?, B :: INTEGER?) :
      | in   | a | b |
      | null | 1 | 2 |
    When executing query:
      """
      CALL test.my.proc(null) YIELD <rename>
      RETURN <a>, <b>
      """
    Then the result should be, in order:
      | <a> | <b> |
      | 1   | 2   |
    And no side effects
    
    Examples:
      | <rename>       | <a> | <b> |
      | a AS c, b AS d | c   | d   |
      | a AS b, b AS d | b   | d   |
      | a AS c, b AS a | c   | a   |
      | a AS b, b AS a | b   | a   |
      | a AS c, b AS b | c   | b   |
      | a AS c, b      | c   | b   |
      | a AS a, b AS d | a   | d   |
      | a, b AS d      | a   | d   |
      | a AS a, b AS b | a   | b   |
      | a AS a, b      | a   | b   |
      | a, b AS b      | a   | b   |
    
```

As illustrated, this makes iterating through a larger number of case very simple.

Similarly a the corresponding failing scenarios can be arrange in a single scenario outline.

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,153 @@
...
+    And there exists a procedure test.my.proc(in :: INTEGER?) :: (A :: INTEGER?, B :: INTEGER?) :
+      | in   | a | b |
+      | null | 1 | 2 |
+    When executing query:
+      """
+      CALL test.my.proc(null) YIELD a, b as a
+      RETURN a
+      """
+    Then a SyntaxError should be raised at compile time: VariableAlreadyBound
+
```
hvub, line 169:
```suggestion
```

tck/features/clauses/call/Call5.feature
```diff
@@ -29,3 +29,153 @@
...
+    When executing query:
+      """
+      CALL test.my.proc(null) YIELD out AS a
+      RETURN a
+      """
+    Then the result should be, in order:
+      | a     |
+      | 'nix' |
+    And no side effects
+
```
hvub, line 62: This scenario seems to be wrongly named. If the scenario is changed to actually do the renaming in the `RETURN` clause, it should be moved to Call6. Otherwise and as it is, the scenarios seems to covered by the scenarios outline I suggest in a comment below and should probably removed (if not changed and moved to Call6).

```suggestion
  Scenario: [2] Rename procedure call output in RETURN clause
    Given an empty graph
    And there exists a procedure test.my.proc(in :: INTEGER?) :: (out :: STRING?):
      | in   | out   |
      | null | 'nix' |
    When executing query:
      """
      CALL test.my.proc(null) YIELD out
      RETURN out AS a
      """
    Then the result should be, in order:
      | a     |
      | 'nix' |
    And no side effects

```

tck/features/clauses/call/Call5.feature
swilly22: Agree order shouldn't have an impact.

tck/features/clauses/call/Call5.feature
swilly22: sounds good to me

hvub:
Looks good to me. Let's kick it over the fence. :-)

PR #491, opened 2021-04-08
Fill feature expressions/pattern/Pattern2

hvub:
This PR aims at providing scenarios for the so far empty feature `expressions/pattern/Pattern2`.


hvub:
* `size(<pattern>)` is deprecated
 * this PR should:
   * Write CIP
   * remove `pattern/Pattern2`
   * rename `pattern/Pattern3` to `pattern/Pattern2`
   * add scenario on `size(<pattern comprehension>)` to `list/List6`
   * add failing scenario on `size(<pattern>)` to `list/List6`

hvub:
Ad "add scenario on `size(<pattern comprehension>)` to `list/List6`":

The scenarios `pattern/Pattern3.[4]`, `pattern/Pattern3.[7]`–`[9]` testing on `size(<pattern comprehension>)`. I will simply move them over to `list/List6`. The scenarios are not picture-prefect and could be improved. Let's not do this now, but rather as a part of improving/completing `list/List6` whenever that comes up.

tck/features/expressions/pattern/Pattern2.feature
```diff
@@ -28,5 +28,212 @@
 
 #encoding: utf-8
 
-Feature: Pattern2 - Number of Pattern Matches
+Feature: Pattern3 - Pattern Comprehension
```
hvub, line 31:
```suggestion
Feature: Pattern2 - Pattern Comprehension
```

tck/features/expressions/pattern/Pattern2.feature
```diff
@@ -28,5 +28,211 @@
...
+             (a)-[:T]->(d)
+      """
+    When executing query:
+      """
+      MATCH (n:A)
+      RETURN [p = (n)-->(:B) | p]
+      """
+    Then the result should be, in any order:
+      | [p = (n)-->(:B) \| p] |
+      | [<(:A)-[:T]->(:B)>]   |
```
Mats-SX, line 69: Since the result column looks rather weird here, I would suggest aliasing it to something
```suggestion
      RETURN [p = (n)-->(:B) | p] AS aToB
      """
    Then the result should be, in any order:
      | aToB                |
      | [<(:A)-[:T]->(:B)>] |
```

tck/features/expressions/pattern/Pattern2.feature
```diff
@@ -28,5 +28,211 @@
...
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A), (b:B)
+      CREATE (a)-[:T]->(b)
+      """
+    When executing query:
+      """
+      MATCH (a:A), (b:B)
+      RETURN [p = (a)-[*]->(b) | p] AS paths
```
Mats-SX, line 82: the scenario name doesn't mention variable length pattern, and the scenario setup doesn't really model something where a variable length causes an interesting effect. Thus, I would suggest using a simple relationship pattern here instead.
```suggestion
      RETURN [p = (a)-->(b) | p] AS paths
```

tck/features/expressions/pattern/Pattern2.feature
```diff
@@ -28,5 +28,211 @@
...
+      """
+    Then the result should be, in any order:
+      | paths               |
+      | [<(:A)-[:T]->(:B)>] |
+    And no side effects
+
+  Scenario: [4] Introducing new node variable in pattern comprehension
```
Mats-SX, line 89:
```suggestion
  Scenario: [4] Introducing a new node variable in pattern comprehension
```

tck/features/expressions/pattern/Pattern2.feature
```diff
@@ -28,5 +28,211 @@
...
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a), (b {name: 'val'})
+      CREATE (a)-[:T]->(b)
```
Mats-SX, line 94: not a comment on this PR per se, but I'm curious if you have some guideline towards the formatting of setup queries?

Concretely, I'm thinking whether there is some stylistic rule that dictates the form you chose rather than these equivalent forms:

```
CREATE (a)-[:T]->(b {name: 'val'})
```
```
CREATE (a), (b {name: 'val'}),
       (a)-[:T]->(b)
```

tck/features/expressions/pattern/Pattern2.feature
```diff
@@ -28,5 +28,211 @@
...
+      """
+    When executing query:
+      """
+      MATCH (n)
+      RETURN [(n)-[:T]->(b) | b.name] AS list
+      """
+    Then the result should be, in any order:
+      | list    |
+      | ['val'] |
+      | []      |
```
Mats-SX, line 104: I think it would add an interesting level to this scenario if we also added another node `c` which is connected like `(b)-[:T]->(c)` and which does _not_ have a `name` property. That would mean the result adds a third row with a `[null]` value, illustrating that the result list size varies not with the presence of the property, but with the cardinality of the pattern.

Mats-SX: I was initially confused because I thought `[]` corresponded to the `a` node not having a name, but it corresponds to the `b` node having no outgoing rels. My suggestion may help another reader.

tck/features/expressions/pattern/Pattern2.feature
```diff
@@ -28,5 +28,211 @@
...
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A), (:A), (:A)
+      CREATE (a)-[:HAS]->()
+      """
+    When executing query:
+      """
+      MATCH (n:A)
+      RETURN count([p = (n)-[:HAS]->() | p]) AS c
```
Mats-SX, line 135: kind of a weird query, but seems OK. anything non-null will add 1 per incoming row, and a pattern comprehension is never null.

Mats-SX: I guess it validates that these are just expressions also when given to aggregators

tck/features/expressions/pattern/Pattern2.feature
```diff
@@ -28,5 +28,211 @@
...
+             (m1)-[:T]->(i1),
+             (m1)-[:T]->(i2)
+      CREATE (n2:X {n: 2}), (m2), (i3:L), (i4:Y)
+      CREATE (n2)-[:T]->(m2),
+             (m2)-[:T]->(i3),
+             (m2)-[:T]->(i4)
+      """
+    When executing query:
+      """
+      MATCH p = (n:X)-->(b)
```
Mats-SX, line 157: can make `b` anonymous
```suggestion
      MATCH p = (n:X)-->()
```

Mats-SX (changes requested):
I had a few suggested changes. Very good work!

tck/features/expressions/pattern/Pattern2.feature
hvub: I added you suggestion. The cardinality beyond 0 and 1 can only be tested correctly by using `size` since the outcome of a pattern comprehension is indeterministic with regards to the order of the elements in the list, as far as I understand things. It is a problem in scenario [1], which I now have changed to not be indeterministic.

hvub:
Thanks @Mats-SX, though in general I have to say, I did not came up with these scenarios, they existed already (in Pattern3). They certainly can be improved. However, that is not the aim of this PR. This PR aimed at providing scenarios for `size(<pattern>)` (originally Pattern2) which turned out to be deprecated. Hence, this PR removes the feature Pattern2 provisioned for `size(<pattern>)` and renames Pattern3 to Pattern2. 

Anyway, some improvements have be made. More can be made at some later stage. At the moment, we try to focus on filling complete empty features.

Mats-SX (approved):

PR #496, opened 2021-06-16
added sub queries scenarios

DvirDukhan:
This PR adds 3 feature file followed by [CIP2015-05-13 Existential subqueries](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-05-13-EXISTS.adoc)

The feature files are located under `useCases` folder and are logically ordered as:
1. Simple Existential Subqueries
2. Full Existential Subqueries
3. Nested Subqueries


tck/features/useCases/existentialSubqueries/ExistentialSubqueries3.feature
```diff
@@ -0,0 +1,71 @@
...
+        MATCH (m) WHERE EXISTS {
+          MATCH (l) WHERE (l)<--(n)-->(m)
+        }
+      }
+      RETURN n
+      """
+    Then the result should be, in any order:
+      | n             |
+      | (:A {prop:1}) |
+    And no side effects
```
hvub, line 71: empty line missing

hvub:
One minor quick observation: The majority of the features are name with noun(s) in singular. Hence, I would prefer to call the new features file `ExistentialSubquery<N>.feature` including the descriptive part, e.g. "ExistentialSubquery3 - Nested existential subquery".

hvub:
Found the problem:

A [SinglePartQuery](https://raw.githack.com/openCypher/openCypher/master/tools/grammar-production-links/grammarLink.html?p=SinglePartQuery) always has to end in a ReturnClause. 🤦🏻‍♂️

Hence,

```
MATCH (n) WHERE EXISTS {
   MATCH (m) WHERE (n)-->(m)
}
```

has to become something like

```
MATCH (n) WHERE EXISTS {
   MATCH (m) WHERE (n)-->(m) RETURN true
}
RETURN n
```

to be syntactically correct. Note the extra ReturnClause on the inner query as well as on the outer query.

The simple version (without MATCH) still does not need and in fact does not even allow a ReturnClause on the inner query.

Hence, 

```
MATCH (n) WHERE EXISTS {
   (m) WHERE (n)-->(m)
}
```

has to become something like

```
MATCH (n) WHERE EXISTS {
   (m) WHERE (n)-->(m)
}
RETURN n
```

BTW: The ReturnClause was also missing on inner queries in some examples of the [CIP](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-05-13-EXISTS.adoc). I have fixed that editorially.

AviAvni:
> Found the problem:
> 
> A [SinglePartQuery](https://raw.githack.com/openCypher/openCypher/master/tools/grammar-production-links/grammarLink.html?p=SinglePartQuery) always has to end in a ReturnClause. 🤦🏻‍♂️
> 
> Hence,
> 
> ```
> MATCH (n) WHERE EXISTS {
>    MATCH (m) WHERE (n)-->(m)
> }
> ```
> 
> has to become something like
> 
> ```
> MATCH (n) WHERE EXISTS {
>    MATCH (m) WHERE (n)-->(m) RETURN true
> }
> RETURN n
> ```
> 
> to be syntactically correct. Note the extra ReturnClause on the inner query as well as on the outer query.
> 
> The simple version (without MATCH) still does not need and in fact does not even allow a ReturnClause on the inner query.
> 
> Hence,
> 
> ```
> MATCH (n) WHERE EXISTS {
>    (m) WHERE (n)-->(m)
> }
> ```
> 
> has to become something like
> 
> ```
> MATCH (n) WHERE EXISTS {
>    (m) WHERE (n)-->(m)
> }
> RETURN n
> ```
> 
> BTW: The ReturnClause was also missing on inner queries in some examples of the [CIP](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-05-13-EXISTS.adoc). I have fixed that editorially.

in the last example there is not diff

AviAvni:
@hvub please approve running workflows

hvub:
The last example has indeed a difference, specifically a ReturnClaus in the outer query.

Lookin at the log. There is one query

```
MATCH (n) WHERE EXISTS {
  MATCH (m) WHERE EXISTS {
    (n)-[]->(m) WHERE n.prop = m.prop
  }
}
```
that lacks exactly the ReturnClause in the outer query. Another one

```
MATCH (n) WHERE EXISTS {
  MATCH (m) WHERE EXISTS {
    MATCH (l) WHERE (l)<--(n)-->(m) RETURN true
  }
}
RETURN n
```

lacks the ReturnClause in the first inner query.

A bush of other query fails the style checker. The error message on those is not the most beautiful but decodable, e.g.

```
[2] Full existential subquery with aggregation *** FAILED ***
        the query of step `When executing query:` in line 57 has either a syntax conforming to the style requirements or the scenario has the @skipStyleCheck tag; expected style: MATCH (n) WHERE exists {
          MATCH (m) WHERE (n)-->(m)
          WITH m, count(*) AS numConnections
          WHERE numConnections = 3
          RETURN true
        }
        RETURN n (MATCH (n) WHERE EXISTS {
          MATCH (m) WHERE (n)-->(m)
          WITH m, count(*) as numConnections
          WHERE numConnections = 3
          RETURN true
        }
        RETURN n,false) did not match the given pattern (ValidateQuery.scala:62)
```

complains about `exists` should be written in capital letter, i.e. `EXISTS`. Same for the `AS` keyword. The error message shows the query twice: first as it is in the scenario and second as it is stylistically expected to be. You can copy both into some diff tool if you can spot the difference right away, which is sometimes hard. For instance:

image

This is from [text-compare.com](https://text-compare.com/). Any other diff tool will certainly do too.

AviAvni:
@hvub workflow need approval again

AviAvni:
@hvub please help with the last error 
```
- [2] Nested full existential subquery *** FAILED ***
23568
        the query of step `When executing query:` in line 60 has either a syntax conforming to the grammar or the scenario has the @skipGrammarCheck tag (Failure(java.lang.AssertionError: ambiguity in query: MATCH (n) WHERE exists {
23569
          MATCH (m) WHERE exists {
23570
            MATCH (l) WHERE (l)<--(n)-->(m) RETURN true
23571
          }
23572
          RETURN true
23573
        }
23574
        RETURN n),false) did not match the given pattern (ValidateQuery.scala:55)
```

hvub:
@AviAvni There is a shortcoming in the generate parser the build is using. This shortcoming makes the parser think very brief patterns such as `(a)<--(b)` in an expression context can be parsed in multiple ways, hence, it complains. Note that other manual written parser do not have this problem. It only effects the oC build.

The parser thinks `(a)<--(b)` could also mean an expression `(a) < -( -(a) ) )` (with extra parentheses for clarity), i.e. "parenthesized `a` is less than the twice negated parenthesized `b`".

The problem is known. No one got around solving it. 

However, the problem is easy to work around in most scenarios, since relationships have a type which can be mentioned in the pattern to break the ambiguity for the parser.

In your case, in ExistentialSubquery3.[2] turning `<--` into `<-[:R]-` is enough:

```
    When executing query:
      """
      MATCH (n) WHERE exists {
        MATCH (m) WHERE exists {
          MATCH (l) WHERE (l)<-[:R]-(n)-->(m) RETURN true
        }
        RETURN true
      }
      RETURN n
      """
```


Putting a label predicate on the first node of the pattern, e.g. `(l:C)<-[:R]-(n)-->(m)` should do the trick, too.

Hope this helps.

AviAvni:
@hvub  thanks for the explanation

tck/features/useCases/existentialSubqueries/ExistentialSubquery2.feature
```diff
@@ -0,0 +1,83 @@
...
+      """
+      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), (a)-[:R]->(:C {prop: 2}), (a)-[:R]->(:D {prop: 3})
+      """
+    When executing query:
+      """
+      MATCH (n) WHERE exists {
+        MATCH (m) WHERE (n)-->(m)
+        RETURN true
+      }
+      RETURN n
```
hvub, line 45: This is not ideal, since the subquery nests a pattern predicate. It is not wrong, but it is a special case. Pattern predicates are in a way the simplest for of an existential subquery, so I'd rather have that in feature ExistentialSubquery3. Here, I would do the following:

```suggestion
      MATCH (n) WHERE exists {
        MATCH (n)-->()
        RETURN true
      }
      RETURN n
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery2.feature
```diff
@@ -0,0 +1,83 @@
...
+      """
+    When executing query:
+      """
+      MATCH (n) WHERE exists {
+        MATCH (m) WHERE (n)-->(m)
+        WITH m, count(*) AS numConnections
+        WHERE numConnections = 3
+        RETURN true
+      }
+      RETURN n
```
hvub, line 66: same as above:

```suggestion
      MATCH (n) WHERE exists {
        MATCH (n)-->(m)
        WITH m, count(*) AS numConnections
        WHERE numConnections = 3
        RETURN true
      }
      RETURN n
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery2.feature
```diff
@@ -0,0 +1,83 @@
...
+
+#encoding: utf-8
+
+Feature: ExistentialSubquery2 - Full existential subquery
+
+  Scenario: [1] Full existential subquery
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), (a)-[:R]->(:C {prop: 2}), (a)-[:R]->(:D {prop: 3})
```
hvub, line 37: Just a formatting thing for better readability:

```suggestion
      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), 
             (a)-[:R]->(:C {prop: 2}), 
             (a)-[:R]->(:D {prop: 3})
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery2.feature
```diff
@@ -0,0 +1,83 @@
...
+    Then the result should be, in any order:
+      | n             |
+      | (:A {prop:1}) |
+    And no side effects
+
+  Scenario: [2] Full existential subquery with aggregation
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), (a)-[:R]->(:C {prop: 2}), (a)-[:R]->(:D {prop: 3})
```
hvub, line 56: Here, I'd add a case to the data that is actually filtered out based on the aggregation. Otherwise, the effect of the aggregation does not really materialize in the result set:

```suggestion
      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), 
             (a)-[:R]->(:C {prop: 2}), 
             (a)-[:R]->(d:D {prop: 3}), 
             (b)-[:R]->(d)
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery2.feature
```diff
@@ -0,0 +1,83 @@
...
+
+  Scenario: [3] Full existential subquery with update clause should fail
+    Given any graph
+    When executing query:
+      """
+      MATCH (n) WHERE exists {
+        MATCH (m) WHERE (n)-->(m)
+        SET m.prop='fail'
+      }
+      RETURN n
```
hvub, line 81: same as above:

```suggestion
      MATCH (n) WHERE exists {
        MATCH (n)-->(m)
        SET m.prop='fail'
      }
      RETURN n
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery3.feature
```diff
@@ -0,0 +1,73 @@
...
+
+#encoding: utf-8
+
+Feature: ExistentialSubquery3 - Nested existential subquery
+
+  Scenario: [1] Nested simple existential subquery
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), (a)-[:R]->(:C {prop: 2}), (a)-[:R]->(:D {prop: 3})
```
hvub, line 37: Just a formatting thing for better readability:

```suggestion
      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), 
             (a)-[:R]->(:C {prop: 2}), 
             (a)-[:R]->(:D {prop: 3})
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery3.feature
```diff
@@ -0,0 +1,73 @@
...
+    Then the result should be, in any order:
+      | n             |
+      | (:A {prop:1}) |
+    And no side effects
+
+  Scenario: [2] Nested full existential subquery
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), (a)-[:R]->(:C {prop: 2}), (a)-[:R]->(:D {prop: 3})
```
hvub, line 58: Just a formatting thing for better readability:

```suggestion
      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), 
             (a)-[:R]->(:C {prop: 2}), 
             (a)-[:R]->(:D {prop: 3})
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery3.feature
```diff
@@ -0,0 +1,73 @@
...
+        RETURN true
+      }
+      RETURN n
+      """
+    Then the result should be, in any order:
+      | n             |
+      | (:A {prop:1}) |
+    And no side effects
+
+  Scenario: [2] Nested full existential subquery
```
hvub, line 54: I would split up this scenario in two, one without and one with pattern predicate.

```suggestion
  Scenario: [2] Nested full existential subquery
    Given an empty graph
    And having executed:
      """
      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), 
             (a)-[:R]->(:C {prop: 2}), 
             (a)-[:R]->(:D {prop: 3})
      """
    When executing query:
      """
      MATCH (n) WHERE exists {
        MATCH (m) WHERE exists {
          MATCH (l)<-[:R]-(n)-[:R]->(m) RETURN true
        }
        RETURN true
      }
      RETURN n
      """
    Then the result should be, in any order:
      | n             |
      | (:A {prop:1}) |
    And no side effects
    
  Scenario: [3] Nested full existential subquery with pattern predicate
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery1.feature
```diff
@@ -0,0 +1,101 @@
...
+
+#encoding: utf-8
+
+Feature: ExistentialSubquery1 - Simple existential subquery
+
+  Scenario: [1] Simple subquery without WHERE clause
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), (a)-[:R]->(:C {prop: 2}), (a)-[:R]->(:D {prop: 3})
```
hvub, line 37: Just a formatting thing for better readability:

```suggestion
      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), 
             (a)-[:R]->(:C {prop: 2}), 
             (a)-[:R]->(:D {prop: 3})
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery1.feature
```diff
@@ -0,0 +1,101 @@
...
+    Then the result should be, in any order:
+      | n             |
+      | (:A {prop:1}) |
+    And no side effects
+
+  Scenario: [2] Simple subquery with WHERE clause
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), (a)-[:R]->(:C {prop: 2}), (a)-[:R]->(:D {prop: 3})
```
hvub, line 55: Here, I'd add a case to the data that is actually filtered out by the inner WHERE clause. Otherwise, the effect of the WHERE clause does not really materialize in the result set:

```suggestion
      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), 
             (a)-[:R]->(:C {prop: 2}), 
             (a)-[:R]->(:D {prop: 3}), 
             (b)-[:R]->(d)
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery1.feature
```diff
@@ -0,0 +1,101 @@
...
+    Then the result should be, in any order:
+      | n             |
+      | (:A {prop:1}) |
+    And no side effects
+
+  Scenario: [3] Simple subquery without WHERE clause, not existing pattern
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), (a)-[:R]->(:C {prop: 2}), (a)-[:R]->(:D {prop: 3})
```
hvub, line 73: Just a formatting thing for better readability:

```suggestion
      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), 
             (a)-[:R]->(:C {prop: 2}), 
             (a)-[:R]->(:D {prop: 3})
```

tck/features/useCases/existentialSubqueries/ExistentialSubquery1.feature
```diff
@@ -0,0 +1,101 @@
...
+      """
+    Then the result should be, in any order:
+      | n |
+    And no side effects
+
+  Scenario: [4] Simple subquery with WHERE clause, not existing pattern
+    Given an empty graph
+    And having executed:
+      """
+      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), (a)-[:R]->(:C {prop: 2}), (a)-[:R]->(:D {prop: 3})
```
hvub, line 90: Just a formatting thing for better readability:

```suggestion
      CREATE (a:A {prop: 1})-[:R]->(b:B {prop: 1}), 
             (a)-[:R]->(:C {prop: 2}), 
             (a)-[:R]->(:D {prop: 3})
```

hvub (changes requested):
Thank you for the recent changes. I gave the PR a more detailed review.

As a global comment: Please move the feature files from `tck/useCases/existentialSubqueries` to `tck/expressions/existentialSubqueries`. Existential subqueries are a form of expressions.

Apart from that, I left a bunch of inline suggestions for slight improvements.

I also have some thoughts on additional scenarios. However, we can add such later. I have noted them in the gaps spreadsheet. Nevertheless, the essences is captured here. 

Apart from what is mentioned, the PR looks good to me. Let's move forward and sense some progress. :-)

hvub (approved):

PR #497, opened 2021-06-18
Fixing YIELD grammar

hvub:
PR this is a reaction to #460.

According to the [CIP on CALL](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2015-06-24-call-procedures.adoc) standalone procedure calls allow `YIELD *` while in-query procedure calls do not allow it.

This PR encodes this rule for `YIELD *` in the grammar. It also adds two scenarios to the TCK that test the two cases of the rule.

In passing, the PR removes `<seq> ... </seq>` tags that are immediately contained in `<production ...> ... </production>` tags. Productions imply a sequence, hence, such `<seq> ... </seq>` tags are superfluous.

### DIFF
--- a/grammar/basic-grammar.xml
+++ b/grammar/basic-grammar.xml
@@ -262,16 +262,14 @@
   </production>
 
   <production name="StringOperatorExpression" rr:inline="true">
-    <seq>
-      <alt>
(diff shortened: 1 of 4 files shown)


thobe (approved):
I think these changes look good.

PR #499, opened 2021-07-20
CIP on grouping keys and aggregation expressions

hvub:
This PR provides CIP that clarifies the semantics of grouping key and aggregation expressions in the WITH and the RETURN clause.

[Rendered view](https://github.com/hvub/openCypher/blob/CIP-on-grouping-keys-and-aggregation-expressions/cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc).

### DIFF
--- a/cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
+++ b/cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
@@ -133,7 +133,7 @@ projects each row in the driving table to `b` minus `a` (as `x`) and the product
 
 The WITH and the RETURN clauses also allow ordering and truncating the driving table, but let's ignore these aspects in this CIP.
 
-=== Beyond baseline semantics -- simple rewrite semantics[[simpleRewrite]]
+=== Simple rewrite semantics[[simpleRewrite]]
 
 Cypher's WITH and the RETURN are syntactically more flexible when the two baseline semantics.
 In particular, they allow mixing aggregation and projection rather freely.
@@ -223,7 +223,7 @@ RETURN x AS x, SUM(sumBC) AS sumBC
 ----
 ====
 
-=== Beyond simple rewrite semantics
+=== Limits of simple rewrite semantics
 
 While this solution works nicely for the considered examples, it is limited.
 Specifically, it only supports aggregation function in expressions of the form `_agg_(_ex_)`.
--- a/cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
+++ b/cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
@@ -10,18 +10,17 @@
 [abstract]
 .Abstract
 --
-
+This CIP clarifies the semantics of grouping and aggregation functionality in the WITH and the RETURN clause of Cypher.
 --
 
 toc::[]
 
 == Background
 
-=== Problem
-
 In this CIP we assume the following baseline semantics as a given and well-defined.
+Further, we consider a simple rewrite semantics for grouping and aggregation, that lays the foundation for the rewrite semantics that is proposed in the <<Proposal>> section of this CIP.
 
-We also assume following driving table in examples:
+We also use following driving table in most of the CIP's examples:
 
 .Example driving table
 |===
@@ -32,7 +31,7 @@ We also assume following driving table in examples:
 |2|3|5
 |===
 
-==== Baseline aggregation
+=== Baseline aggregation
 
 Cypher allows grouping and aggregating (form here on just referred to as _aggregation_) the driving table in the WITH and in the RETURN clause.
 
@@ -83,7 +82,7 @@ Grouped the driving table by `a` and computes the sum of all `c` as `sumC` for e
 |===
 ====
 
-==== Baseline projection
+=== Baseline projection
 
 The WITH and the RETURN clause also allow projecting the driving table including the computation of new columns (in database theory, this is called extended projection).
 
@@ -134,7 +133,7 @@ projects each row in the driving table to `b` minus `a` (as `x`) and the product
 
 The WITH and the RETURN clauses also allow ordering and truncating the driving table, but let's ignore these aspects in this CIP.
 
-==== Beyond baseline semantics -- simple rewrite semantics[[simpleRewrite]]
+=== Beyond baseline semantics -- simple rewrite semantics[[simpleRewrite]]
 
 Cypher's WITH and the RETURN are syntactically more flexible when the two baseline semantics.
 In particular, they allow mixing aggregation and projection rather freely.
@@ -188,7 +187,7 @@ Further, for a set of <ProjectionItem>s _L_ = {_p_~1~, _p_~2~, ... _p~N~_},
 * Let `PROJ(_L_)` be `PROJ(_p_~1~), PROJ(_p_~2~), ..., PROJ(_p~N~_)` and
 * Let `AGGR(_L_)` be `AGGR(_p_~1~), AGGR(_p_~2~), ..., AGGR(_p~N~_)`
 
-With this the semantics of `RETURN _L_` can be defined as effectively equivalent to
+With this, `RETURN _L_` can be defined as effectively equivalent to
 
 [source, cypher, subs="quotes"]
 ----
@@ -196,7 +195,7 @@ WITH PROJ(_L_)
 RETURN AGGR(_L_)
 ----
 
-Analogously, the semantics of `WITH _L_` can be defined as effectively equivalent to
+Analogously, `WITH _L_` can be defined as effectively equivalent to
 
 [source, cypher, subs="quotes"]
 ----
@@ -204,16 +203,18 @@ WITH PROJ(_L_)
 WITH AGGR(_L_)
 ----
 
+Let's call this the _simple rewrite semantics_ for the WITH and RETURN clause.
+
 .Simple rewrite semantics
 ====
-With this semantics, the RETURN clause in <<Q3>>
+With the simple rewrite semantics, the RETURN clause in <<Q3>>
 
 [source, cypher]
 ----
 RETURN b-a AS x, SUM(b*c) AS sumBC`)
 ----
 
-is defined as effectively equivalent to
+is effectively equivalent to
 
 [source, cypher]
 ----
@@ -222,9 +223,7 @@ RETURN x AS x, SUM(sumBC) AS sumBC
 ----
 ====
 
-Let's call this the _simple rewrite semantics_ for the WITH and RETURN clause.
-
-==== Beyond simple rewrite semantics
+=== Beyond simple rewrite semantics
 
 While this solution works nicely for the considered examples, it is limited.
 Specifically, it only supports aggregation function in expressions of the form `_agg_(_ex_)`.
@@ -293,13 +292,15 @@ _None. -- This proposal does not propose and net-new syntax._
 
 === Semantics
 
-The proposed semantics defined as a rewrite to the baseline semantics (similar to <<simpleRewrite,simple rewrite semantics>> discussed above).
+The proposed grouping and aggregation semantics is defined as a rewrite to the baseline semantics (similar to <<simpleRewrite,simple rewrite semantics>> discussed above).
 The proposed semantics does not cover all queries and hence implies a syntax restriction to rule out queries that are not covered.
 We discuss the rewrite and the syntax restriction in the follow two subsections.
 
 ==== Rewrite
 
-For a <ProjectionItem> _p_ = `_postEx_ AS _al_`, let _AGG_(_p_) be the set of (sub-)expressions _aggex_ of the form _agg_(_preEx_).
+For an expression _ex_, let _AGG_(_ex_) be the set of (sub-)expressions _aggex_ of the form _agg_(_preEx_).
+
+For a <ProjectionItem> _p_ = `_postEx_ AS _al_`, let _AGG_(_p_) be the set of (sub-)expressions _aggex_ of the form _agg_(_preEx_), i.e. _AGG_(_p_) = _AGG_(_postEx_).
 
 The set of <ProjectionItem>s _L_ is split according to _AGG_(_p_) in two cases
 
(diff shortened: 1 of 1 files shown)


cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,779 @@
...
+----
+====
+
+.Rewrite semantics
+====
+The RETURN clause in <<Q4>>
+
+[source, cypher]
+----
+RETURN a AS a, (a + SUM(b*c) - MIN(c)) * 2 AS foo
```
LinneaAndersson, line 357: I would have liked to see an example where the expression that contains an aggregation also contains a  non-leaf grouping expression. For example:

```
RETURN a.x, a.x + MIN(a.y)
```

How would this be rewritten with the rules?

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,779 @@
...
+. `MATCH (a)-[]->(b) RETURN a, a.x + count(b.y)`
+* The sub-expression `a` is a grouping key.
+
+. `MATCH (a)-[]->(b) RETURN a, size(keys(a)) + count(b.y)`
+* The sub-expression `a` is a grouping key.
+
+. `MATCH (x) RETURN x.a, x.b, x.c, x.a + x.b + count(x) + x.c`
+* The sub-expressions `x.a`, `x.b`, and `x.c` are grouping keys.
+
+. `MATCH (a) RETURN a.x + 1, a.x + 1 + count(a.x)`
```
LinneaAndersson, line 573: Should this really be valid? Didn't we say it would be too confusing since we couldn't allow `MATCH (a) RETURN a.x + 1, count(a.x) + a.x + 1 `?

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,779 @@
...
+* The sub-expression `$x` is a parameter.
+
+. `RETURN 1 + count($x) + $x * 7 + sum($x) + 'cake'`
+* The sub-expressions `1`, `2`, and `'cake'` are constants.
+* The sub-expression `$x` is a parameter.
+
+. `MATCH (a) RETURN a.x, 1 + count(a.x)`
+* The sub-expression `1` is a constant.
+
+. `MATCH (a) RETURN a.x, a.x + count(a.x)`
```
LinneaAndersson, line 558: As mentioned in another comment as well. Didn't we end up with the rules:

```
In an expression that contains aggregation, the leaves of expressions must be one of:
An aggregation
A literal
A Parameter
A variable - ONLY IF that variable is also a projection expression on its own (e.g. the n in RETURN n AS myNode, n.value + count(*))
```

In this expression `a` is the leaf and it is _not_  a projection expression on its own

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,779 @@
...
+
+_GROUPING_KEYS_(_L_) is {`b-a`, `x`, `c`}.
+
+====
+
+For an expression _ex_ and a projection list _L_, let _CONSTANT_(_ex_, _L_) hold
+
+* If _ex_ is either
+** A aggregation function, i.e. of the form `_agg_(_subEx_)`,
+** A grouping key, i.e. _ex_ ∈ _GROUPING_KEYS_(_L_),
```
LinneaAndersson, line 449: Didn't we update the rules to:

```
In an expression that contains aggregation, the leaves of expressions must be one of:
An aggregation
A literal
A Parameter
A variable - ONLY IF that variable is also a projection expression on its own (e.g. the n in RETURN n AS myNode, n.value + count(*))
```

LinneaAndersson (commented):
Great job! 

I just have a few questions for you regarding the rules

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
hvub:
```
RETURN a.x AS foo, a.x + MIN(a.y) AS bar
```

what it should become

```
WITH a.x AS foo, a.y AS bar_1
WITH foo AS foo, MIN(bar_1) AS bar_1
RETURN foo AS foo, foo + bar_1  AS foo
```

You have a good point, in that currently the rewrite does not do that yet.

But it seems easy to fix. The definition of _POST_PROJ_(_p_) needs adjust, so that is also replaces all subexpression that are grouping keys with their alias.

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
hvub: That is related to what is written in the caveats section. From the openCypher grammar standpoint there is no left-deep understanding to chained numeric additions. So there is no hook to hang such a restriction on.

I do not like this either, but I think it goes beyond this CIP to change the grammar in that regard. We could escape the problem by saying in the CIP that the CIP assumes a left-deep parse tree. However, the CLG notes did not indicate that this assumption was part of the decision, hence, I did not felt mandated to do it. 

I rather have a CLG discussion on that issues -- that is why I wrote the caveat part.

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
hvub: Well, initially I thought so too.

However, attached to one CLG card is implementation report pdf attached to this card
https://trello.com/c/56tgXSFP/326-grouping-key-confusion-when-combining-aggregating-and-non-aggregating-values
(It isn't a google doc unfortunately for some reason, so it is unclear if there is newer version of it.)

This document lists exactly this query `MATCH (a) RETURN a.x, a.x + count(a.x)` as number 7 in the section called "Ok (Not deprecated)".

Going pure leaves would certainly be cleaner. But it would als deprecate even more query. Not everyone would be happy about that as you can imaging.

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
hvub: See comment above.

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
LinneaAndersson: I guess we need to bring this up, to make it more clear how it should be. 

(On the 6th of August I wrote, on the same card as you just linked, a couple of examples where `MATCH (a) RETURN a.x, a.x + count(a.x)` was one of the non-allowed queries)

Mats-SX (commented):
I noticed I had a pending review of this since a while back. Apologies for the delayed response. My review is mainly editorial.

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,779 @@
...
+.Abstract
+--
+This CIP clarifies the semantics of grouping and aggregation functionality in the WITH and the RETURN clause of Cypher.
+--
+
+toc::[]
+
+== Background
+
+In this CIP we assume two baseline semantics as a given and well-defined.
```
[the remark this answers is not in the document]
hvub:
```suggestion
In this CIP we assume two baseline semantics as a given and as well-defined.
```

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,779 @@
...
+
+Analogously, `WITH _p_~1~, _p_~2~, ..., _p~N~_` can be defined as effectively equivalent to
+
+[source, cypher, subs="quotes"]
+----
+WITH _PROJ_(_p_~1~), _PROJ_(_p_~2~), ..., _PROJ_(_p~N~_)
+WITH _AGGR_(_p_~1~), _AGGR_(_p_~2~), ..., _AGGR_(_p~N~_)
+----
+
+We call this the _simple rewrite semantics_ for the WITH and RETURN clause.
```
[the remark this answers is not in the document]
hvub: Thanks. :-)

hvub:
Thanks @Mats-SX for the editorial fixes.

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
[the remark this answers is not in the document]
Mats-SX: I'm reacting to "as a given". I'm not familiar with the use of "a given" in that way. Something can be given, then it is "we assume it to be given" or "we assume it is given" or "we assume it as given". I'm not a linguist, but "as a given" reads odd to me.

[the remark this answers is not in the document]
hvub: I apparently got confused about what is old and what is new. I remove the "a" and the second "as".

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,779 @@
...
+.Abstract
+--
+This CIP defines the semantics of grouping and aggregation functionality in the WITH and the RETURN clause of Cypher.
+--
+
+toc::[]
+
+== Background
+
+In this CIP we assume two baseline semantics as a given and as well-defined.
```
hvub, line 20:
```suggestion
In this CIP we assume two baseline semantics as given and well-defined.
```

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,833 @@
...
+
+. `MATCH (a)-[c]->(b) WITH a, c, {dim: properties(b)} AS b RETURN a, b.dim.x, sum(c.p) + b.dim.x`
+* The sub-expression `b.dim.x` is not a recognized grouping key.
+
+=== Valid orderings
+
+The following clauses exhibit valid row orderings.
+For each example we list why it is valid.
+
+. `RETURN 1 + count(\*) AS x ORDER BY x`
```
LinneaAndersson, line 680:
```suggestion
. `RETURN 1 + count(*) AS x ORDER BY x`
```

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,833 @@
...
+The following clauses exhibit invalid row orderings.
+For each example we list why it is invalid.
+
+. `RETURN 1 + count(\*) ORDER BY 1 + count(*)`
+* The expression `1 + count(*)` is not a grouping key.
+
+. `MATCH (a) RETURN a.x + 1, a.x + 1 + count(a.x) ORDER BY a.x + 1`
+* The sub-expression `a.x + 1` is a recognized grouping key.
+
+. `MATCH (a) RETURN a.x + 1, a.x + 1 + count(a.x) ORDER BY a.x + 2`
```
LinneaAndersson, line 735: Just as a note. This (and some of the examples below) doesn't run at all. It gives an exception: `In a WITH/RETURN with DISTINCT or an aggregation, it is not possible to access variables declared before the WITH/RETURN: a `. But I believe that might be another discussion and not related to this  CIP.

LinneaAndersson (approved):
Very well written 🎉

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
```diff
@@ -0,0 +1,833 @@
...
+
+The proposed grouping and aggregation semantics is defined as a rewrite to the baseline semantics (similar to <<Simple rewrite semantics>> discussed above).
+The proposed semantics does not cover all syntactically possible queries and hence requires a syntax restriction to prohibit queries that are not covered.
+We discuss the <<Rewrite>> and the <<Syntax restriction>> in the following two subsections.
+We simplify this discussion by ignoring row ordering and pagination as well as omitted aliases.
+Subsequently, we give a separate and brief consideration of how to these aspects fit into the proposal, cf. <<Row ordering and pagination>> and <<Omitted aliases>>, respectively.
+
+==== Rewrite
+
+For an expression _ex_, let _AGG_(_ex_) be the set of aggregating (sub-)expressions, i.e. (sub-)expressions _aggex_ of the form _agg_(_preEx_).
```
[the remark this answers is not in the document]
hvub: It stands to a subexpression in _AGG_(_ex_). However, you are right in that the symbol is not needed since not used further down. So let's simply remove it:

```suggestion
For an expression _ex_, let _AGG_(_ex_) be the set of aggregating (sub-)expressions, i.e. (sub-)expressions of the form _agg_(_preEx_).
```

cip/1.accepted/CIP2021-07-07-Grouping-keys-and-aggregation-expressions.adoc
hvub: Well, it is listed under "Invalid orderings" so `MATCH (a) RETURN a.x + 1, a.x + 1 + count(a.x) ORDER BY a.x + 2` is supposed to fail. 

However, on the one above, `MATCH (a) RETURN a.x + 1, a.x + 1 + count(a.x) ORDER BY a.x + 1`: The explanation text misses a "not", I think. It should say "The sub-expression `a.x + 1` is **not** a recognized grouping key." This, I will have to fix.

PR #500, opened 2021-07-23
Add TCK error characteristics wildcards

hvub:
With this PR scenarios can specify error characteristics (error type, the error phase, the error details) with a wildcard in the ExpectedError step of a scenario.

The wildcards are used in a scenario as follows:

- **Any error type:** `Then a Error should be raised at ...: ...`
- **Any error phase:** `Then a ... should be raised at any time: ...`
- **Any error detail:** `Then a ... should be raised at ...: *`

Where a wildcard is given, the TCK API will not test the respective error characteristics on the error returned by an implementation for the respective scenario step. In other words, a scenario with an error characteristics wildcard cannot fail on this error characteristics in this ExpectedError step. The scenario may still fail 
- on other steps
- on other error characteristics that are not specified with a wildcard, or 
- because the implementation fails produce an error at all.

All scenarios that is at least one error characteristics wildcard, get tagged with `@WildcardErrorDetails` so that they are easily identifiable.

### DIFF
--- a/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
+++ b/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
@@ -267,14 +267,18 @@ object CypherTCK {
       scenarioSteps
     }.toList
     val (name, number) = parseNameAndNumber(nameAndNumber)
-    val tagsInferred = {
-      if (steps.exists {
+    val tagsInferred = tags ++ Set(TCKTags.NEGATIVE_TEST, TCKTags.WILDCARD_ERROR_DETAILS).filter {
+      case TCKTags.NEGATIVE_TEST => steps.exists {
         case ExpectError(_, _, _, _) => true
         case _ => false
-      })
-        tags + TCKTags.NEGATIVE_TEST
-      else
-        tags
+      }
+      case TCKTags.WILDCARD_ERROR_DETAILS => steps.exists {
+        case ExpectError(TCKErrorTypes.ERROR, _, _, _) => true
+        case ExpectError(_, TCKErrorPhases.ANY_TIME, _, _) => true
+        case ExpectError(_, _, TCKErrorDetails.ANY, _) => true
+        case _ => false
+      }
+      case _ => false
     }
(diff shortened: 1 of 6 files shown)


tools/tck-api/src/test/resources/org/opencypher/tools/tck/Foo.feature
```diff
@@ -47,6 +47,38 @@ Feature: Foo
       """
     Then a SyntaxError should be raised at compile time: UnknownFunction
 
+  Scenario: Fail with any type
+    Given an empty graph
+    When executing query:
+      """
+      RETURN foo()
+      """
+    Then a Error should be raised at compile time: UnknownFunction
```
tobias-johansson, line 56:
```suggestion
    Then an Error should be raised at compile time: UnknownFunction
```

tools/tck-api/src/test/resources/org/opencypher/tools/tck/Foo.feature
```diff
@@ -47,6 +47,38 @@ Feature: Foo
...
+      """
+    Then a SyntaxError should be raised at compile time: *
+
+  Scenario: Fail with any type and any detail at any type
+    Given an empty graph
+    When executing query:
+      """
+      RETURN foo()
+      """
+    Then a Error should be raised at any time: *
```
tobias-johansson, line 80:
```suggestion
    Then an Error should be raised at any time: *
```

tobias-johansson (approved):
:+1:

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
```diff
@@ -267,14 +267,18 @@ object CypherTCK {
       scenarioSteps
     }.toList
     val (name, number) = parseNameAndNumber(nameAndNumber)
-    val tagsInferred = {
-      if (steps.exists {
+    val tagsInferred = tags ++ Set(TCKTags.NEGATIVE_TEST, TCKTags.WILDCARD_ERROR_DETAILS).filter {
+      case TCKTags.NEGATIVE_TEST => steps.exists {
         case ExpectError(_, _, _, _) => true
```
hedengran, line 272:
```suggestion
        case _: ExpectError => true
```

hedengran (approved):

PR #503, opened 2021-08-13
Add support for Scala cross building for 2.12 and 2.13

mnd999:
* Add two maven profiles, `scala-212` and `scala-213` to build for Scala 2.12 or Scala 2.13, keeping the existing default of Scala 2.12. 
* Due to collection library incompatibilities between 2.12 and 2.13, tck-api requires a slightly different classpath for each, so I've added `scala_2.12` and `scala_2.13` additional source directories for this module. 
* Added additional maven plugins for the additional src directory and rewriting the Scala versions in the .poms
* Changed the output of the side effects reporting to make the order it displays additions and deletions consistent. This was previously unordered and behaving differently on 2.12 and 2.13, causing a test failure.
* Upgrade some dependencies to be compatible with 2.12 and 2.13

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,6 @@
     <dep.maven.source.plugin.version>3.2.1</dep.maven.source.plugin.version>
     <dep.maven.surefire.plugin.version>2.22.2</dep.maven.surefire.plugin.version>
     <dep.scala.maven.plugin.version>3.4.4</dep.scala.maven.plugin.version>
-    <dep.scala.version>${scala.version}</dep.scala.version>
     <dep.scalatest.version>3.2.2</dep.scalatest.version>
     <dep.scalatest.runner.version>0.1.8</dep.scalatest.runner.version>
     <dep.xmlgraphics-commons.version>2.4</dep.xmlgraphics-commons.version>
@@ -54,7 +53,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.rootdir>${project.basedir}</project.rootdir>
     <scala.binary.version>2.12</scala.binary.version>
-    <scala.version/>
+    <scala.version>2.12.13</scala.version>
     <test.runner.jvm/>
     <test.runner.jvm.settings.additional/>
     <test.runner.jvm.settings>
@@ -88,7 +87,7 @@
       <dependency>
         <groupId>org.scala-lang</groupId>
         <artifactId>scala-library</artifactId>
-        <version>${dep.scala.version}</version>
+        <version>${scala.version}</version>
       </dependency>
 
       <dependency>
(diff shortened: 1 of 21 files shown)


pom.xml
```diff
@@ -38,7 +39,7 @@
     <dep.maven.source.plugin.version>3.2.1</dep.maven.source.plugin.version>
     <dep.maven.surefire.plugin.version>2.22.2</dep.maven.surefire.plugin.version>
     <dep.scala.maven.plugin.version>3.4.4</dep.scala.maven.plugin.version>
-    <dep.scala.version>2.12.10</dep.scala.version>
+    <dep.scala.version>${scala.version}</dep.scala.version>
```
hvub, line 42: Is this indirection from `dep.scala.version` to `scala.version` necessary? It seems odd to have two properties for the same thing.

pom.xml
```diff
@@ -53,6 +54,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.rootdir>${project.basedir}</project.rootdir>
     <scala.binary.version>2.12</scala.binary.version>
```
hvub, line 56: Why do `scala.binary.version` and `dep.cask.version` have "default" here, which gets overwritten by the profile, while `scala.version` does not have such a default? Is there a specific reason? 

If not, I guess, I would prefer to either have defaults for all profile-related properties OR for none of them.

tools/tck-api/src/main/scala_2.13/org/opencypher/tools/tck/api/events/Collections.scala
```diff
@@ -0,0 +1,50 @@
+/*
```
boggle, line 1: Why is this added by this PR? Also, tangentially and at a glance, the whole thing feels hacky.

hvub: The respective methods disappeared in Scala 2.13 without real replacement.

pom.xml
mnd999: Maybe not, I can try removing it.

pom.xml
mnd999: The [docs for the cross-build-plugin](https://github.com/kelnos/scala-cross-maven-plugin#usage) suggest it's a good idea to keep scala.binary.version to keep IntelliJ happy. I suspect the other is an oversight, I will try and remove it.

hvub (approved):
I think I understand what it does. It looks good to me. I approve it.

One open questions for me: What happens during release build (on Teamcity)? Does it produce artifacts for both profiles and releases them to [Maven Central](https://search.maven.org/search?q=g:org.opencypher%20tck)?

Thanks.

mnd999:
@hvub I haven't touched the release process, as I don't really know it. But yes, ideally we should publish artifacts for both. Once this is merged maybe we can figure this out between us.

PR #504, opened 2021-09-01
Add missing results order check in tck scenarios

smietana:
TCK scenarios should check order if query ends with ORDER BY


thobe (approved):
Looks good and sensible to me! Thank you!

PR #505, opened 2021-09-24
Grammar toolkit

thobe:
This adds the ability to read and process WG3 grammar files.
It also adds the ability to work with grammar "projects", which allows defining a set of inputs (that can relate to one another) and a set of outputs derived from those inputs. The "projects" notion should be able to replace scripts that run multiple tools over grammars to generate a set of outputs.
There is also a feature where a grammar can be defined as a patch on top of another grammar. This is useful when working on a proposed change in order to be able to test the whole new grammar in total while keeping a neat record of the changes made.

I will say that the state of this work is not my finest. It is a bit "work in progress". In particular it is lacking in testing, documentation and could do with quite a bit of refactoring to simplify things. But it does work, and it does not break things that have worked before.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -128,6 +128,12 @@
       </dependency>
 
       <dependency>
+        <groupId>io.cucumber</groupId>
+        <artifactId>cucumber-gherkin-vintage</artifactId>
+        <version>${dep.cucumber.version}</version>
+      </dependency>
+
+      <dependency>
         <groupId>org.apache.xmlgraphics</groupId>
         <artifactId>batik-svggen</artifactId>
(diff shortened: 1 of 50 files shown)


thobe:
I have no idea why that build issue has not been seen before. It should have been unrelated to these changes. It is fixed now though. I also discovered and fixed another issue in the process that was introduced by this code and only manifested when running from the command line (the resources were not included in the jar).

hvub:
I have discussed the principles of the changes with @thobe in persons. Based on that and given there are no hard dependencies on this code, I think we can merge the PR.

PR #509, opened 2021-10-29
Add TCK step for creating CSV files

hedengran:


tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Scenario.scala
```diff
@@ -203,6 +213,10 @@ case class Scenario(categories: List[String], featureName: String, number: Optio
           throw ScenarioFailedException(msg, error)
       }
     }
+
+    def withParameters(additionalParameters: Map[String, CypherValue]): ScenarioExecutionContext = {
```
hvub, line 217: I wonder if this name captures better what the function is doing:

```suggestion
    def addParameters(additionalParameters: Map[String, CypherValue]): ScenarioExecutionContext = {
```

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Scenario.scala
```diff
@@ -96,12 +96,22 @@ case class Scenario(categories: List[String], featureName: String, number: Optio
 
           case (ctx, RegisterProcedure(signature, table, _)) =>
             ctx.graph match {
-              case support: ProcedureSupport =>
-                support.registerProcedure(signature, table)
+              case procedureSupport: ProcedureSupport =>
+                procedureSupport.registerProcedure(signature, table)
               case _ =>
             }
             Right(ctx)
```
hvub, line 103: To keep the code structure parallel with cvs support:

```suggestion
              case procedureSupport: ProcedureSupport =>
                Right(procedureSupport.registerProcedure(signature, table))
              case _ =>
                Right(ctx)
            }
```

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Scenario.scala
```diff
@@ -158,7 +168,7 @@ case class Scenario(categories: List[String], featureName: String, number: Optio
             else Right(ctx)
 
           case (ctx, Parameters(ps, _)) =>
-            Right(ctx.copy(parameters = ps))
+            Right(ctx.withParameters(ps))
```
hvub, line 171: To be consistent with my other comment below:

```suggestion
            Right(ctx.addParameters(ps))
```

tools/tck-api/src/test/resources/org/opencypher/tools/tck/Foo.feature
```diff
@@ -39,6 +39,25 @@ Feature: Foo
       | 1 |
     And no side effects
 
+  Scenario: Reading a CSV
+    Given an empty graph
+    And there exists a CSV file with URL as $param, with rows:
+      | 1 | 'a' |
```
hvub, line 45: I would test the whole roundtrip (multiple comments):

```suggestion
    And there exists a CSV file with URL as $param, with rows:
      | i | txt |
      | 1 | 'a' |
```

These data tables also expect headers, I think.

tools/tck-api/src/test/resources/org/opencypher/tools/tck/Foo.feature
```diff
@@ -39,6 +39,25 @@ Feature: Foo
...
+
+    When executing query:
+    """
+    LOAD CSV WITH HEADERS from '$param' AS row
+    RETURN $list AS res
+    """
+    Then the result should be, in any order:
+      | res   |
+      | []    |
+      | []    |
```
hvub, line 58: I would test the whole roundtrip (multiple comments):

```suggestion
    RETURN row.txt AS res
    """
    Then the result should be, in any order:
      | res |
      | 'a' |
      | 'b' |
```

tools/tck-api/src/test/scala/org/opencypher/tools/tck/TckTest.scala
```diff
@@ -52,7 +52,7 @@ class TckTest extends AnyFunSpec with Assertions {
     }
   }
 
-  private object FakeGraph extends Graph with ProcedureSupport {
+  private object FakeGraph extends Graph with ProcedureSupport with CsvFileCreationSupport {
```
hvub, line 55: I would test the whole roundtrip (multiple comments):

```suggestion
  private object FakeGraph extends Graph with ProcedureSupport with CsvFileCreationSupport {
    var cvsData = CypherValueRecords.empty
```

tools/tck-api/src/test/scala/org/opencypher/tools/tck/TckTest.scala
```diff
@@ -63,12 +63,17 @@ class TckTest extends AnyFunSpec with Assertions {
...
+          StringRecords(List("res"), List(Map("res" -> "[]"), Map("res" -> "[]")))
         case ExecQuery =>
           StringRecords(List("1"), List(Map("1" -> "1")))
       }
     }
 
     override def registerProcedure(signature: String, values: CypherValueRecords): Unit =
       ()
+
+    override def createCSVFile(contents: CypherValueRecords): String = "dummy/path.csv"
```
hvub, line 77: I would test the whole roundtrip (multiple comments):

```suggestion
    override def createCSVFile(contents: CypherValueRecords): String = {
      cvsData = contents
      "dummy/path.csv"
    }
```

tools/tck-api/src/test/scala/org/opencypher/tools/tck/TckTest.scala
```diff
@@ -63,12 +63,17 @@ class TckTest extends AnyFunSpec with Assertions {
           CypherValueRecords.empty
         case ExecQuery if query.contains("foo()") =>
           ExecutionFailed(SYNTAX_ERROR, COMPILE_TIME, UNKNOWN_FUNCTION)
+        // assert that csv path parameter is not overwritten by additional parameters
+        case ExecQuery if query.contains("LOAD CSV") && params.keySet.equals(Set("param", "list")) =>
+          StringRecords(List("res"), List(Map("res" -> "[]"), Map("res" -> "[]")))
```
hvub, line 68: I would test the whole roundtrip (multiple comments):

```suggestion
          StringRecords(List("res"), cvsData.rows.map(r => Map("res" -> r("txt").toString)))
```

hvub (changes requested):
Thanks for the PR. In principle, it looks good to me. In a nutshell, I suggest

 * a slightly better function name than `withParameters` and
 * an extension of the test, such that it check that the data table is actually correctly forwarded to the `CsvFileCreationSupport` by the `CsvFile` step.

Thanks.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Scenario.scala
hedengran: `.registerProcedure()` returns a unit, so  this is not possible, unfortunately.

hvub (changes requested):
Looks very nice. One of your code comments made me think about scenarios declaring duplicate parameter names. That would be confusing. Hence I think, we should prevent in the with an appropriate step in the TCK integrity test. A comment provides more detail.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Scenario.scala
```diff
@@ -96,12 +96,22 @@ case class Scenario(categories: List[String], featureName: String, number: Optio
 
           case (ctx, RegisterProcedure(signature, table, _)) =>
             ctx.graph match {
-              case support: ProcedureSupport =>
-                support.registerProcedure(signature, table)
+              case procedureSupport: ProcedureSupport =>
+                Right(procedureSupport.registerProcedure(signature, table))
               case _ =>
             }
             Right(ctx)
```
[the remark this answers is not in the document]
hedengran: Sorry for the half-edit. I understand the purpose. But `Right(procedureSupport.registerProcedure(signature, table))` is not possible since `registerProcedure()` returns `Unit`. So it will have to be the same way it was to begin with.

tools/tck-api/src/test/scala/org/opencypher/tools/tck/TckTest.scala
```diff
@@ -87,13 +88,21 @@ class TckTest extends AnyFunSpec with Assertions with Matchers {
           CypherValueRecords.empty
         case ExecQuery if query.contains("foo()") =>
           ExecutionFailed(SYNTAX_ERROR, COMPILE_TIME, UNKNOWN_FUNCTION)
+        // assert that csv path parameter is not overwritten by additional parameters
```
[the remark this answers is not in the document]
hedengran: Nice, good point! I'll add this.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
```diff
@@ -350,6 +350,23 @@ case class RegisterProcedure(signature: String, values: CypherValueRecords, sour
   }
 }
 
+case class CsvFile(parameter: String, table: CypherValueRecords, source: io.cucumber.core.gherkin.Step) extends Step {
```
arnefischereit, line 353: Maybe also leave a comment on this case class on what `parameter` means.

tools/tck-integrity-tests/src/test/scala/org/opencypher/tools/tck/ValidateSteps.scala
```diff
@@ -135,6 +137,20 @@ trait ValidateSteps extends AppendedClues with Matchers with OptionValues with V
       case _ => succeed
     }
 
+    withClue("scenario does not declare conflicting parameters") {
```
arnefischereit, line 140: Did you add a test for this?

tools/tck-integrity-tests/src/test/scala/org/opencypher/tools/tck/ValidateSteps.scala
```diff
@@ -135,6 +137,20 @@ trait ValidateSteps extends AppendedClues with Matchers with OptionValues with V
...
+      val parameters = steps.collect {
```
arnefischereit, line 141: I guess this would need to be `flatMap` or the like, so that we do not end up with a `List[List[Parameter]]`.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Scenario.scala
```diff
@@ -205,6 +216,10 @@ case class Scenario(categories: List[String], featureName: String, number: Optio
           throw ScenarioFailedException(msg, error)
       }
     }
+
+    def addParameters(additionalParameters: Map[String, CypherValue]): ScenarioExecutionContext = {
+      this.copy(parameters = this.parameters ++ additionalParameters)
```
arnefischereit, line 221: Shouldn't this do a check that the key sets of `parameters` and `additionalParameters` are disjoint?

arnefischereit (changes requested):
When you introduce the check in `ValidateSteps`, I think it would make sense to also add a test for that.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/constants/TCKStepDefinitions.scala
```diff
@@ -59,6 +59,9 @@ object TCKStepDefinitions {
   val INSTALLED_PROCEDURE = """^there exists a procedure (.+):$"""
   val installedProcedureR = INSTALLED_PROCEDURE.r
 
+  val CSV_FILE = "^there exists a CSV file with URL as \\$(.*), with rows:$"
```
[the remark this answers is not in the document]
hedengran: What do you mean? Quote the parameter?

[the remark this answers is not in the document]
arnefischereit: Sorry, ignore this point.
I misread this at first that one should provide the URL like `with URL as '/home/arne/file.csv'` but that is not the case and I forgot to remove this comment later on.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Scenario.scala
hedengran: That sanity-check already lives in `ValidateSteps` now?

tools/tck-integrity-tests/src/test/scala/org/opencypher/tools/tck/ValidateSteps.scala
```diff
@@ -135,6 +137,20 @@ trait ValidateSteps extends AppendedClues with Matchers with OptionValues with V
       case _ => succeed
     }
 
+    withClue("Scenario declare conflicting parameter ") {
```
arnefischereit, line 140:
```suggestion
    withClue("Scenario declares conflicting parameter ") {
```

arnefischereit (approved):
Nice test :+1:
Consider whether you want to document `CsvFile`, otherwise go ahead to merge this.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Scenario.scala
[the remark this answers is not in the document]
hvub: You are right.

hvub (approved):

PR #514, opened 2021-11-05
Fix a bug in UnaryAddSubtractExpression

Mats-SX:
The bug allowed arbitrarily many occurrences of `+` or `-` operators.
Only a single operator is allowed as part of this expression.
Since a literal expression also allows a preceding `-` character (for
a negative number literal), `--1` for example is also allowed.


hvub:
Thanks @Mats-SX.

Looks good to me. Can be merged.

PR #516, opened 2021-11-22
Restructure Scala Dependencies

mnd999:
For the 2.13 cross build to work the Scala dependencies cannot be in the top level .pom as this will not work in a dependency hierarchy unless everything has a Scala version. To fix, move all the Scala dependencies under the scala modules so that the non-Scala modules can stay as they are. tck-reporting has direct Scala dependencies so no choice there.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -115,13 +115,7 @@
         <scope>import</scope>
       </dependency>
 
-      <dependency>
-        <groupId>io.cucumber</groupId>
-        <artifactId>cucumber-scala_${scala.binary.version}</artifactId>
-        <version>${dep.cucumber.version}</version>
-      </dependency>
-
-      <dependency>
+       <dependency>
         <groupId>io.cucumber</groupId>
         <artifactId>cucumber-junit</artifactId>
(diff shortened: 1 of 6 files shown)


hvub:
1) So you essentially avoid the dependency management functionality of the parent pom for scala dependencies, right?

2) You turned `tools/tck-reporting` into being scala binary version coded. The module itself contains Java code only. It the scala binary version coding need because it has scala dependencies?

3) One thing I notice in passing is in `tools/tck-reporting/pom.xml` which this PR code fix along the way is:

```
        <dependency>
            <!-- dependency of tck-api listed here for intellij to get the scala version right -->
            <groupId>com.lihaoyi</groupId>
            <artifactId>fastparse_${scala.binary.version}</artifactId>
            <version>${dep.fastparse.version}</version>
        </dependency>
```

should not be needed. It is something, I have added and later removed while fiddling, but missed to remove it in this particular pom. Similar poms like `tools/tck-integrity-tests/pom.xml` do not have it anymore.

hedengran (approved):
Looks good, and resolved the issues when I built locally.

PR #517, opened 2021-11-23
Revert incorrect update to test assertion

hedengran:
* This assertion was initially correct
* After some discussion we thought that it was wrong
* We now understand why it initially was correct (we missed that 'false AND null -> false')


tobias-johansson (approved):
:+1:

hvub:
Looks perfect. Thanks.

PR #528, opened 2022-02-11
Some CypherLists that are equal do not have the same hashcode

mnd999:
On Scala 2.13, MurMurHash3 has changed it's contract to include the prefix in the hash calculation by default. This means that a CypherOrderedList and a CypherUnorderedList with the same content can be equal, but have different hashcodes. This breaks the equals / hashcode contract and is responsible for a TCK failure in Neo4j.
Unfortunately, the mechanism for including or excluding the prefix does not exist in Scala 2.12, so to ensure we can still cross build I have to pull out Scala API version specific calls to MurMurHash3.

### DIFF
--- a/tools/tck-api/src/main/scala/org/opencypher/tools/tck/values/CypherValue.scala
+++ b/tools/tck-api/src/main/scala/org/opencypher/tools/tck/values/CypherValue.scala
@@ -27,8 +27,6 @@
  */
 package org.opencypher.tools.tck.values
 
-import scala.util.hashing.MurmurHash3
-
 object CypherValue {
   def apply(s: String, orderedLists: Boolean = true): CypherValue = {
     val parser = new CypherValueParser(orderedLists)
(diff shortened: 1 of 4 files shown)


hvub (approved):
Hi @mnd999,

your changes look good to me.

Somewhat to the side but still related:

`CypherOrderedList.equals` looks somewhat strange, in particular this line

```
case o: CypherUnorderedList => o == this
```

https://github.com/opencypher/openCypher/pull/528/files#diff-0c12dbaaef29f1dccb4adf49ddf5f01116095ef418b86e284efa0375b0aada6bL117

The commutative case in `CypherUnorderedList.equal` seems to make more sense:

```
case other: CypherOrderedList =>
      other.elements.sorted(CypherValue.ordering) == elements.sorted(CypherValue.ordering)
```

I guess that works, because `o == this` essentially executes the commutative case.

However, I wonder if that is good coding style. I find it confusing at first sight.

Anyhow, it is just a personal observation. I leave it to you to decide if you want to do something about it or not.

mnd999:
@hvub I added a comment to that effect. I don't know if that is better.

hvub:
@mnd999 Oh yes. That is helpful enough without changing the actual code. Very good. Thanks.

Issue #361, opened 2019-04-04
Issues with whitespace definition


thobe:
Neither Java's `Character.isWhitespace(int)`, or `Character.isSpaceChar(int)`, or the unicode [`[:Whitespace:]`](https://unicode.org/cldr/utility/list-unicodeset.jsp?a=%5B%3AWhitespace%3A%5D) specification treats [`\u180E` (MONGOLIAN VOWEL SEPARATOR)](https://unicode.org/cldr/utility/character.jsp?a=180E) as a whitespace.

Yet the openCypher grammar considers this a whitespace character, why?
https://github.com/opencypher/openCypher/blob/346aa0d6ec0d500c84bcc08edb633ba9d8815e73/grammar/basic-grammar.xml#L781

Furthermore the definition of whitespace in the openCypher grammar does not consider [`\u0085` (NEXT LINE)](https://unicode.org/cldr/utility/character.jsp?a=0085) or [`\u00A0` (NBSP)](https://unicode.org/cldr/utility/character.jsp?a=00A0) to be whitespace, while they are part of the unicode [`[:Whitespace:]`](https://unicode.org/cldr/utility/list-unicodeset.jsp?a=%5B%3AWhitespace%3A%5D) specification and considered whitespace by Java's `Character.isSpaceChar(int)`, why?

thobe:
I came across this difference when looking at why the `whitespace` production rules spelled out all whitespace characters individually instead of just referencing the unicode [`[:Whitespace:]`](https://unicode.org/cldr/utility/list-unicodeset.jsp?a=%5B%3AWhitespace%3A%5D) specification. So I investigated the difference.

The conclusion of this exercise is that apart from [`\u0085` (NEXT LINE)](https://unicode.org/cldr/utility/character.jsp?a=0085), the grammar includes all characters of the unicode [`[:Whitespace:]`](https://unicode.org/cldr/utility/list-unicodeset.jsp?a=%5B%3AWhitespace%3A%5D) specification, and additionally includes [`\u001C` (FILE SEPARATOR)](https://unicode.org/cldr/utility/character.jsp?a=001C), [`\u001D` (GROUP SEPARATOR)](https://unicode.org/cldr/utility/character.jsp?a=001D), [`\u001E` (RECORD SEPARATOR)](https://unicode.org/cldr/utility/character.jsp?a=001E), and [`\u001F` (UNIT SEPARATOR)](https://unicode.org/cldr/utility/character.jsp?a=001F).

[...]

PR #530, opened 2022-03-18
Update basic-grammar.xml

hvub:
This PR adresses #361 by implementing @thobe suggestion.


hvub:
Unfortunately, is wasn't as simple as I have hoped. :-(

PR #531, opened 2022-03-23
Stop using deprecated things in TCK tests

Lojjs:
The TCK tests are using syntax which have been deprecated and will be removed in Neo4j 5.0:

- Octal number syntax starting with 0 instead of 0o
- Coercion of lists to boolean


tck/features/expressions/literals/Literals7.feature
```diff
@@ -177,7 +177,7 @@ Feature: Literals7 - List
     Given any graph
     When executing query:
       """
-      RETURN [1, -2, 00, 71034856] AS literal
+      RETURN [1, -2, 0o0, 71034856] AS literal
```
Lojjs, line 180: Given the name of this test, I wonder if the list contained an octal number by mistake/typo and it should just be 0?

tck/features/expressions/pattern/Pattern1.feature
```diff
@@ -38,7 +38,7 @@ Feature: Pattern1 - Pattern predicate
       """
     When executing query:
       """
-      MATCH (n) WHERE (n)-[]->() RETURN n
+      MATCH (n) WHERE NOT isEmpty((n)-[]->()) RETURN n
```
Lojjs, line 41: More or less this entire file was testing patterns which were coerced to boolean. So perhaps we want to remove it fully instead or change the last few tests which are supposed to fail to also use isEmpty?

tck/features/expressions/literals/Literals7.feature
hedengran: Yeah I'm guessing that someone forgot that leading 0 used to mean octal, and thought `00 = 0`. However, after the deprecation cycle, `00` is actually treated as an integer I'd assume? So maybe this test can be left unchanged?

tck/features/expressions/pattern/Pattern1.feature
hedengran: I think the tests still serve a purpose. However, I agree that the negative tests in the end of the file should also be updated to include `isEmpty`.

hedengran (commented):
Looks good! One question though: What's the policy on negative tests in TCK? I'm wondering if we should explicitly test that the replaced syntax is _not_ accepted, as well? And for octals, possibly also that we don't accept uppercase o (`0O123`).

tck/features/expressions/literals/Literals7.feature
hvub: Not necessarily. It demonstrates that the different integer literal formats can be mixed in a list. It may have been more obvious with a non-zero hex number like 0xA4C (something with a letter in it). If you want change it that respect, do not hold yourself back, it would be a win.

tck/features/clauses/match-where/MatchWhere4.feature
```diff
@@ -60,8 +60,8 @@ Feature: MatchWhere4 - Non-Equi-Joins on variables
       """
       MATCH (a), (b)
       WHERE a.id = 0
-        AND (a)-[:T]->(b:TheLabel)
-        OR (a)-[:T*]->(b:MissingLabel)
+        AND NOT isEmpty((a)-[:T]->(b:TheLabel))
+        OR NOT isEmpty((a)-[:T*]->(b:MissingLabel))
```
hvub, line 64: This seems a deliberate decision against `EXISTS { ... }`, which looks like the more straightforward replacement.

```
      MATCH (a), (b)
      WHERE a.id = 0
        AND (a)-[:T]->(b:TheLabel)
        OR (a)-[:T*]->(b:MissingLabel)
        AND EXISTS { (a)-[:T]->(b:TheLabel) }
        OR EXISTS { (a)-[:T*]->(b:MissingLabel) }
      RETURN DISTINCT b
```

Any particular reason or rational behind that?

tck/features/expressions/literals/Literals7.feature
Lojjs: That is a good question. We haven't removed this in Neo4j yet. But I would have assumed that we will throw an error for all cases that used to be deprecated incl. 00

tck/features/expressions/pattern/Pattern1.feature
hvub: No. I think all your changes on naked pattern in `WHERE` go beyond what was decided, far as I am aware of. Naked pattern where not banned completely. They are still allowed in contexts there clearly a boolean is expected and the evaluate to a boolean in these context.

This [CIP2015-05-13 EXISTS](https://github.com/openCypher/openCypher/blob/master/cip/1.accepted/CIP2015-05-13-EXISTS.adoc#212-syntax-rules) summarizes this (the link goes directly to the crucial section). The was written to document the decision made by in the CLG and was also reviewed by the CLG down the line, as far as I remember things.

hvub (changes requested):
Thank you @Lojjs for taking care of this. 

The changes regarding octals are straightforward and perfectly good.

I disagree with the changes on naked patterns in `WHERE`, though. I think, all these scenario do not need a change. They are still valid to my knowledge. Please, see [CIP2015-05-13 EXISTS](https://github.com/openCypher/openCypher/blob/master/cip/1.accepted/CIP2015-05-13-EXISTS.adoc#212-syntax-rules).

tck/features/clauses/match-where/MatchWhere4.feature
Lojjs: Oh this is nicer, I just didn't realise that you could rewrite it like this.

tck/features/expressions/pattern/Pattern1.feature
Lojjs: Hmm, the description in this card for the deprecations https://trello.com/c/AjC65NYS/121-deprecate-coercion-of-lists-to-boolean agrees with you. But what I did was to hack Neo4j to throw errors instead of warnings on deprecations and check which TCK tests failed. And all usages here looked like they was deprecated in the code. I will investigate a bit more...

hvub:
> Looks good! One question though: What's the policy on negative tests in TCK? 

There is not policy on negative test. It is a judgment call. There are necessarily infinite number of things that can make a  negative test. In extreme, is "foo bar" being not a valid Cypher query worth a negative test?

I not explicit thought about this until now. I guess, a behavior could be expected to be possible or syntax combinations the arise from the language's general composition rules or grammar definition but are forbidden warrant a negative test.

> I'm wondering if we should explicitly test that the replaced syntax is _not_ accepted, as well? And for octals, possibly also that we don't accept uppercase o (`0O123`).

Whether my vague guideline give above applies here is a good question.

In the end, every negative test is work. If someone is willing to do the work on a negative test it provides some backing its importance.

hvub:
> The changes regarding octals are straightforward and perfectly good.

One more thing that jumps to mind. Have you checked the [oC grammar](https://github.com/opencypher/openCypher/blob/master/grammar/basic-grammar.xml#L572-L575). I guess that needs to be change accordingly.

The grammar should become something like this, I reckon:

```
  <production name="OctalInteger" oc:lexer="true">
    <literal value="0o"/>
    <repeat min="1"><non-terminal ref="OctDigit"/></repeat>
  </production>
```

If you could include that in the PR, it would be great.

Lojjs:
> Looks good! One question though: What's the policy on negative tests in TCK? I'm wondering if we should explicitly test that the replaced syntax is _not_ accepted, as well? And for octals, possibly also that we don't accept uppercase o (`0O123`).

We might want to use a newer TCK release also for 4.x, e.g. if we find some bug in it. As the old syntax will only be removed in 5.0, I think it can be a bit cumbersome with a negative test for this case.

tck/features/expressions/literals/Literals7.feature
```diff
@@ -177,7 +177,7 @@ Feature: Literals7 - List
     Given any graph
     When executing query:
       """
-      RETURN [1, -2, 00, 71034856] AS literal
+      RETURN [1, -2, 0o77, 0xA4C, 71034856] AS literal
```
hedengran, line 180: You need to update the result assertion as well here.

hedengran (changes requested):
Did you run the updated tests against the Neo4j Cypher implementation? I think it would be nice to test run all tests as a sanity check.

tck/features/expressions/literals/Literals7.feature
hvub: Splendid.

hvub (approved):
Excellent. Good to go. 👍 Thanks you very much.

hedengran (approved):

PR #535, opened 2022-07-11
Update expected error msg for illegal aggregations in order by

LinneaAndersson:


hvub (approved):
I agree the with change.

PR #537, opened 2022-07-19
Update scenario to avoid illegally introducing variable in pattern predicate

hedengran:
### Feature "List6 - List size", scenario "Fail for `size()` on pattern predicates"

Examples 1 to 7 use pattern predicates that introduce variables, which is forbidden (see below). https://github.com/opencypher/openCypher/blob/b0ea6c0de107cf0a9ff27d395adcb07d0baf492a/tck/features/expressions/pattern/Pattern1.feature#L197-L221

To avoid returning a `UndefinedVariable` (in adhearance with above scenario), I suggest we simply avoid introducing new variables in this scenario.

### DIFF
--- a/tck/features/expressions/list/List6.feature
+++ b/tck/features/expressions/list/List6.feature
@@ -95,6 +95,7 @@ Feature: List6 - List size
     Given any graph
     When executing query:
       """
+      MATCH (a), (b), (c)
       RETURN size(<pattern>)
       """
     Then a SyntaxError should be raised at compile time: UnexpectedSyntax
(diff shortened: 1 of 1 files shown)


hvub (approved):
Well, it is a question of which error takes precedence, I guess. An `UnexpectedSyntax` error is semantically correct, too. Unfortunately, we do not have the capability (yet) of listing multiple acceptable errors. I am not going to introduce such capability now. An error wildcard would be another solution, but would make the scenario a little less specific.

In the end, however, the benefit of testing that patterns with new variables are also rejected here is relatively low. Hence, I agree with the PR. Please merge. Thanks.

PR #541, opened 2022-08-30
Remove g4 ressources from antlr demo

hvub:
The PR make modifications to the Antlr parser demo, such that the generate Cypher.g4 file does not reside as a resource in the repository anymore. The resource is removed and the demo changed to look for the file in the current working directory.

Issue #539, opened 2022-08-10
Invalid syntax


howard-gleason:
Thanks for the recent grammar updates that have fixed other syntax errors.  The following Cypher query is valid, however, openCypher throws a parse error
```
MATCH path=()-[]->()
RETURN relationships(path)[0].name
```
In order to get openCypher to parse it, parentheses have to be added:
```
MATCH path=()-[]->()
RETURN (relationships(path)[0]).name
```
Can or will this syntax be accepted into openCypher?

jacobfriedman:
On this note, would it be possible to include the generated .EBNF syntax in this repo as well?

howard-gleason:
@jacobfriedman 
> On this note, would it be possible to include the generated .EBNF syntax in this repo as well?

Milestone artifacts are posted here: https://opencypher.org/resources/

Generate your own by following: https://github.com/opencypher/openCypher/tree/master/grammar

jacobfriedman:
Would it be possible to include the file here so one doesn't have to regenerate it and search for diffs every time the syntax is updated?

PR #542, opened 2022-08-30
Chaining of non arithmetic operators

hvub:
**This PR is based on #541.**

This PR proposes changes to the openCypher grammar for allowing chaining of non-arithmetic operators as a reaction to issue #539. With the PR, the grammar allows for instance 

```
RETURN foo()[1]:F, foo().x.y[1][4].name:F
```

as valid syntax, which was not the case before.

The PR also adjusts [CIP2021-08-10 Operator precedence](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2021-08-10-Operator-precedence.adoc) ([rendered view](https://github.com/hvub/openCypher/blob/chaining-of-non-arithmatic-operators/cip/1.accepted/CIP2021-08-10-Operator-precedence.adoc)) to the changed grammar.


nadja-muller (approved):
It looks like this solves the problem described in the issue to me :+1: 

As a sidenote, comparing it with Cypher (Postfix1 in the code), I think there are some differences:
- We additionally allow listSlices, e.g. 1..2 in the same hierarchy as listOperatorExpression and PropertyLookups. I assume this is part of openCypher.
- We allow node & relationship expressions instead of only NodeLabels.

jonaskello:
Just ran into the issue that `list[0].myprop` will not parse properly, but `(list[0]).myprop` works. From what I understand this PR will fix that. Is there a timeline for when it will be merged/release?

PR #543, opened 2022-09-14
Update for zoneinfo 2022b

mnd999:
Update to reflect timezone offset change to Europe/Stockholm in recent zoneinfo file. Arguably, it may be better to just remove this test, I would also be open to that.


nadja-muller (approved):

mnd999:
We may want to hold off on merging this until all the JVM vendors catch up.

PR #544, opened 2022-09-22
Fix test assertions of labels function

loveleif:
We don't guarantee any ordering of the array returned by the `labels` function, right? These assertions causes failures in  https://github.com/neo-technology/neo4j/pull/17164.


hvub (approved):
Looks good. Please move forward. Thanks.

PR #545, opened 2022-09-23
Add TCK expectation: the result should be, in order (ignoring element order for lists)

smietana:
- Add TCK expectation: the result should be, in order (ignoring element order for lists)
- Do not assert on list elements order in [12] Aggregation of named paths scenario


hvub (approved):
Looks good. On a recent other review I thought the ordered result with unordered lists is missing. Now here it is. 😀

PR #546, opened 2022-09-26
Fix bug in node comparison

loveleif:
Fix a bug where comparison of nodes was incorrect because of labels appearing in different order.

Causes failures in https://github.com/neo-technology/neo4j/pull/17164.

### DIFF
--- a/tools/tck-api/src/test/scala/org/opencypher/tools/tck/values/CypherValueTest.scala
+++ b/tools/tck-api/src/test/scala/org/opencypher/tools/tck/values/CypherValueTest.scala
@@ -126,17 +126,30 @@ class CypherValueTest extends AnyFunSuite with Matchers {
 
     CypherOrderedList(nodeList1) should equal(CypherOrderedList(nodeList2))
     CypherOrderedList(nodeList2) should equal(CypherOrderedList(nodeList1))
-    CypherUnorderedList(nodeList1) should equal(CypherUnorderedList(nodeList2))
-    CypherUnorderedList(nodeList2) should equal(CypherUnorderedList(nodeList1))
     CypherOrderedList(nodeList1) should equal(CypherUnorderedList(nodeList2))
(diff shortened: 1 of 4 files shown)


tools/tck-api/src/test/scala/org/opencypher/tools/tck/values/CypherValueTest.scala
```diff
@@ -107,4 +109,47 @@ class CypherValueTest extends AnyFunSuite with Matchers {
...
+    CypherOrderedList(nodeList1) should equal(CypherOrderedList(nodeList2))
+    CypherOrderedList(nodeList2) should equal(CypherOrderedList(nodeList1))
+    CypherUnorderedList(nodeList1) should equal(CypherUnorderedList(nodeList2))
+    CypherUnorderedList(nodeList2) should equal(CypherUnorderedList(nodeList1))
+    CypherOrderedList(nodeList1) should equal(CypherUnorderedList(nodeList2))
+    CypherOrderedList(nodeList2) should equal(CypherUnorderedList(nodeList1))
+    CypherUnorderedList(nodeList1) should equal(CypherOrderedList(nodeList2))
+    CypherUnorderedList(nodeList2) should equal(CypherOrderedList(nodeList1))
+
+    CypherUnorderedList(nodeList1.reverse) should equal(CypherUnorderedList(nodeList2))
```
hvub, line 136: Would an unordered list1 in reverse order be equal to order list2 or not. We be good to add these test cases.

hvub (approved):
Makes sense. I left one comment about some additional test cases. Feel free to proceed if theses have been added. Thanks.

tools/tck-api/src/test/scala/org/opencypher/tools/tck/values/CypherValueTest.scala
loveleif: The way it's designed it should be equal, added some more test cases. Thanks!

PR #547, opened 2022-10-02
Use COMPARE token for defining comparison operators in PartialComparisonExpression rule

1sand0s:
Signed-off-by: 1sand0s <1sand0sardpi@gmail.com>

Using a `COMPARE` token to define the list of possible comparison operators before their usage in the `oC_PartialComparisonExpression` production rule makes it possible to easily identify which operator was used in the parsed Cypher query (by fetching the terminal node COMPARE).

### DIFF
--- a/tools/grammar-demo/src/main/resources/Cypher.g4
+++ b/tools/grammar-demo/src/main/resources/Cypher.g4
@@ -383,13 +383,10 @@ FALSE : ( 'F' | 'f' ) ( 'A' | 'a' ) ( 'L' | 'l' ) ( 'S' | 's' ) ( 'E' | 'e' )  ;
 oC_ListLiteral
            :  '[' SP? ( oC_Expression SP? ( ',' SP? oC_Expression SP? )* )? ']' ;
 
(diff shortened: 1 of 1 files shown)


arnefischereit (changes requested):
Hi @1sand0s,

Thank you for raising a pull request for improving openCypher.

The problem with your approach is that the lexer will now recognize all equality signs (`=`) as `COMPARE`, while it is used in some other contexts as well (such as `SET`, path assignment, pattern comprehension).

Please also note that [we have since](https://github.com/opencypher/openCypher/commit/38b5e39f9ec61adb63752f1137c2c6840cdc27be) removed `Cypher.g4`, as it is a generated resource. If you think, you can fix the problem of the lexer while keeping your use case, please apply these changes to [basic-grammar.xml](https://github.com/opencypher/openCypher/blob/master/grammar/basic-grammar.xml) instead.

Regards,

Arne

arnefischereit:
Hi @1sand0s,

We have not heard from you in a while. I will now close this issue but please feel free to open a new issue when you want to continue on this work.

Thanks again for your intent to contribute to openCypher.

Regards,

Arne

PR #548, opened 2022-10-18
Support expected errors with side effects

loveleif:
Enables asserting on side effects in failing queries. This is useful for `CALL ... IN TRANSACTIONS` queries, that will fail but also have side effects.

### DIFF
--- a/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
+++ b/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
@@ -46,6 +46,8 @@ import java.net.URI
 import java.nio.charset.StandardCharsets
 import java.nio.file._
 import java.util
+import scala.annotation.tailrec
+import scala.collection.mutable.ListBuffer
 import scala.jdk.CollectionConverters._
 import scala.util.Failure
 import scala.util.Success
@@ -257,7 +259,7 @@ object CypherTCK {
                     If this is a custom error, then disable this validation with tag "${TCKTags.ALLOW_CUSTOM_ERRORS}"""")
             }
           }
-          List(expectedError, SideEffects(source = step).fillInZeros)
+          List(expectedError)
 
         // And
         case noSideEffectsR() => List(SideEffects(source = step).fillInZeros)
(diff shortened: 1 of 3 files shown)


hvub (changes requested):
The changes make sense in principle. However, from the scenario syntax point of view it is not obvious why additional syntax is needed. This seems to be only done because the steps are processed one-by-one `CypherTCK.toScenario`. 

I would rather prefer that a scenario does not have to ", with side effects:", because "Then an Error should be ... And the side effects should be. ..." already says this sufficiently. 

In `CypherTCK.toScenario` such scenarios can be detected by a one step look ahead. The look ahead can be implement, for example, by replacing 

```scala
val steps = pickle.getSteps.asScala.flatMap { step =>
```

with

```scala
    val stepsAsScala = pickle.getSteps.asScala.toSeq
    val successorSteps = stepsAsScala.tail.map(Some(_)) :+ None
    val stepsAsScalaWithSuccessor = stepsAsScala.zip(successorSteps)

    val steps = stepsAsScalaWithSuccessor.flatMap { case (step, successorStep) =>
```

then `successorStep` can be used to decide in `case expectErrorR(errorType, time, detail) => ...` whether to generate a `SideEffects(source = step).fillInZeros` or not.

I think, that given us a cleaner scenario syntax.

loveleif:
> I would rather prefer that a scenario does not have to ", with side effects:", because "Then an Error should be ... And the side effects should be. ..." already says this sufficiently.

@hvub smart! I also didn't like it but didn't thought about that possibility, thanks!

loveleif:
Forced pushed an improved solution to this branch.

hvub (approved):

hvub:
I have looked at this again — commit to do the requested change myself — and found that it is not needed. 

Since `insertSideEffectsOnExpectError` sets the `source` of the inserted `SideEffects` steps to the `source` of the preceding `ExpectError` step and `ExpectError` steps always come with `StepType.THEN`, the inserted `SideEffects` steps fit the check in: 
https://github.com/opencypher/openCypher/blob/master/tools/tck-integrity-tests/src/test/scala/org/opencypher/tools/tck/ValidateSteps.scala#L110-L113


Bottom line: This PR can be merged.

PR #550, opened 2022-11-10
Update to use n as the grouping key, instead of m in Exists Aggregation Acceptance test

gem-neo4j:
This test is currently incorrect, when m is the grouping key, the max numConnections can be is 2, and not 3


hvub (approved):
I fully agree with the observation and the solution. Thank you. Please merge.

PR #551, opened 2023-01-31
Fix bug in unordered rows comparison

loveleif:
Fix bug in `CypherValueRecords.equalsUnordered` where rows with equal hash code could generate false negatives.

### DIFF
--- a/tools/tck-api/src/test/scala/org/opencypher/tools/tck/api/CypherValueRecordsTest.scala
+++ b/tools/tck-api/src/test/scala/org/opencypher/tools/tck/api/CypherValueRecordsTest.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015-2022 "Neo Technology,"
+ * Copyright (c) 2015-2023 "Neo Technology,"
  * Network Engine for Objects in Lund AB [http://neotechnology.com]
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tools/tck-api/src/test/scala/org/opencypher/tools/tck/api/CypherValueRecordsTest.scala
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) 2015-2022 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
```
loveleif, line 7: The build fails because missing license, but I copied this from a neighbouring file, what's wrong?

hvub: It fails on any single license header. I have updated the license header just now. For your new file this should fix it:

```suggestion
/*
 * Copyright (c) 2015-2023 "Neo Technology,"
 * Network Engine for Objects in Lund AB [http://neotechnology.com]
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
 * You may obtain a copy of the License at
```

The alternative route is to rebase to the new master — which has my license header updates — and re-copy the header.

hvub (approved):
Excellent. Thanks for the PR. I have nothing to add. :-)

Please adjust the license headers as shown in the suggestion so it passes the CI and merge. Thanks.

PR #553, opened 2023-02-14
Do not silently fail on scenario initialization queries

hvub:
This PR modifies the TCK API such that an error in an initialization query cause a scenario to fail. Before, such errors where silently swallowed by the API, which makes it hard to debug scenarios that have mistakes in their initialization queries.


hedengran (approved):
Very nice!

PR #557, opened 2023-04-11
Fix bug in comparison of unordered lists

loveleif:
Fix bug in Cypher unordered list comparison. 

For example, assertions like this

```
Then the result should be (ignoring element order for lists):
    | foo                 |
    | ['Bob', 'Carl']     |
    | ['Ada', 'Danielle'] |
    | ['Danielle']        |
    | ['Carl']            |
    | []                  |
```

would fail even if the results were correct:

```
...failed with message: 
Expected (in any order of rows):
| foo |
| ['Bob', 'Carl'] |
| ['Ada', 'Danielle'] |
| ['Danielle'] |
| ['Carl'] |
| [] |
Actual:
| foo |
| ['Carl', 'Bob'] |
| ['Ada', 'Danielle'] |
| ['Danielle'] |
| ['Carl'] |
| [] |
```

Trello: https://trello.com/c/m36sEKjr


mnd999 (approved):
Looks good, I like that we test that equals is symmetrical a lot better than before. One small nit comment.

PR #558, opened 2023-05-09
Test-fix: expect-error step also tests errorType and phase

sherfert:


hvub (approved):
What a dumb bug. Thanks for fixing it.

PR #566, opened 2023-10-10
Relax error type assertions

hedengran:


hvub (approved):
I reckon you doing the minimal thing here. That is find at this point. 

More things need to happen in the future to get the house in order. That is not the goal now. 

Nevertheless, this RP is a little step in the right direction. Please merge.

PR #573, opened 2024-05-16
license header update 2024

mnd999:


hvub (approved):
Thanks.

hvub:
What's the reason the build fails on license check on scala 2.12, though? 🤔

hvub:
Oh... it seems to complain that the grammar xml files do not have a license header.

hvub:
Yes, that's it. The /grammar/*.xml files miss the license header update. You will want to update these five file, too. Presumably needs to be done by hand. If the build is green, feel free to click the merge button. 🙂

mnd999:
Oops yeah, I was rebasing my Scala 3 PR to see if the plugin update helped (it doesn't) and then hit this, so I was skipping the build for those parts since they don't work with Java 17. Should be okay now :hand_with_index_finger_and_thumb_crossed:

PR #576, opened 2024-07-01
Update tck tests

JoaquimGouveia:
In connection with the new GQL status standard I added support for two types of error tests: 
1 - Test which relies on the GQL status code of the exception. 
2 - Test which relies on both the GQL status code and the message of the exception.

### DIFF
--- a/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
+++ b/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
@@ -260,7 +260,8 @@ object CypherTCK {
             }
           }
           List(expectedError)
-
+        case expectErrorWithGQLCodeR(gqlCode) => List(ExpectErrorWithGQLCode(gqlCode, step))
+        case expectErrorWithGQLCodeAndMessageR(gqlCode, message) => List(ExpectErrorWithGQLCodeAndMessage(gqlCode, message, step))
         // And
         case noSideEffectsR() => List(SideEffects(source = step).fillInZeros)
         case sideEffectsR()   => List(SideEffects(parseSideEffectsTable, step).fillInZeros)
@@ -437,6 +438,12 @@ case class ExpectResult(expectedResult: CypherValueRecords, source: io.cucumber.
     val state = Seq(expectedResult, PickleStep(source), sorted)
     state.map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
   }
+}
+
+case class ExpectErrorWithGQLCode(gqlCode: String, source: io.cucumber.core.gherkin.Step) extends Step {
+}
+
+case class ExpectErrorWithGQLCodeAndMessage(gqlCode: String, message: String, source: io.cucumber.core.gherkin.Step) extends Step {
 }
 
 case class ExpectError(errorType: String, phase: String, detail: String, source: io.cucumber.core.gherkin.Step) extends Step {
--- a/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
+++ b/tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
@@ -247,8 +247,8 @@ object CypherTCK {
         case expectSortedResultR()         => List(ExpectResult(parseTable(), step, sorted = true))
         case expectResultUnorderedListsR() => List(ExpectResult(parseTable(orderedLists = false), step))
         case expectSortedResultUnorderedListsR() => List(ExpectResult(parseTable(orderedLists = false), step, sorted = true))
-        case expectErrorR(errorType, time, detail) =>
-          val expectedError = ExpectError(errorType, time, detail, step)
+        case expectErrorWithLegacyDetailR(errorType, time, detail) =>
+          val expectedError = ExpectErrorWithLegacyDetail(errorType, time, detail, step)
           if (shouldValidate) {
             expectedError.validate() match {
               case None => // No problem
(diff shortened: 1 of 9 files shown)


tools/tck-api/src/test/resources/org/opencypher/tools/tck/Foo.feature
```diff
@@ -67,6 +67,22 @@ Feature: Foo
...
+      """
+    Then an error with GQL code fooCode should be raised
+
+  Scenario: Fail with code and message
+    Given an empty graph
+    When executing query:
+      """
+      return foo()
+      """
+    Then an error with GQL code fooCode should be raised with message fooMessage
```
loveleif, line 84: Asserting on the complete message can be difficult (it often contains the complete query string, and that might be long and in parameterized scenarios (examples) it is different between runs). So I suggest to make it `...with message containing`.

Or perhaps have syntax like:

```
Then an error with GQL code fooCode should be raised with message containing:
  """Blabla"""
```

tools/tck-api/src/test/resources/org/opencypher/tools/tck/Foo.feature
```diff
@@ -67,6 +67,22 @@ Feature: Foo
       """
     Then a SyntaxError should be raised at compile time: UnknownFunction
 
+  Scenario: Fail with code
+    Given an empty graph
+    When executing query:
+      """
+      RETURN foo()
+      """
+    Then an error with GQL code fooCode should be raised
```
loveleif, line 76: Perhaps we should quote the code? `Then an error with GQL code 'fooCode' should be raised` and/or `Then an error with GQL code "fooCode" should be raised`.

hvub: It does not seem to matter. The code do not contain spaces, I think, so they are visually clearly delineated. I was thinking about a colon before the message, like this: `Then an error with GQL code fooCode should be raised with message: bla bla bla foo bla`. This seems optically helpful.

tools/tck-api/src/test/resources/org/opencypher/tools/tck/Foo.feature
hvub: Or you could have:

```
Then an error with GQL code fooCode should be raised with message matching: /regex/
```

The question is what is actually needed. In the oC scenarios, I would advocate for just checking the code, as the message is something vendor specific.

If we want to check the precise message for the product, than I am not opposing these means. But maybe there is another add this so that the scenarios do not have to be duplicated. 

Maybe the TCK API can be given a mapping of code to message regex (by the  Neo4j TCK adapter, in our case). When evaluating the error step of a scenario the TCK API checks is it has a mapping for the code. If it has none, it does not check the message. If it has one, it test the message too. How about that?

This would allow us to test messages for consistency without unnecessarily detailing the oC TCK with messages.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Result.scala
```diff
@@ -82,4 +82,4 @@ object CypherValueRecords {
   }
 }
 
-case class ExecutionFailed(errorType: String, phase: String, detail: String, exception: Option[Throwable] = None)
+case class ExecutionFailed(errorType: String, gqlCode: String, message: String, phase: String, detail: String, exception: Option[Throwable] = None)
```
hvub, line 85: Can we split this into two classes? Something like this:

```
sealed trait ExecutionFailed
case class ExecutionFailedWithLegacyDetail(errorType: String, phase: String, detail: String, exception: Option[Throwable] = None) extends ExecutionFailed
case class ExecutionFailedWithGQLCode(gqlCode: String, message: String, exception: Option[Throwable] = None) extends ExecutionFailed
```

Maybe that is too harsh of an API change. But on the other hand, it makes clearer where this is supposed to go. The GQL code is the means to get rid of errortype, phase, and detail.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
```diff
@@ -439,6 +440,12 @@ case class ExpectResult(expectedResult: CypherValueRecords, source: io.cucumber.
   }
 }
 
+case class ExpectErrorWithGQLCode(gqlCode: String, source: io.cucumber.core.gherkin.Step) extends Step {
+}
+
+case class ExpectErrorWithGQLCodeAndMessage(gqlCode: String, message: String, source: io.cucumber.core.gherkin.Step) extends Step {
+}
+
 case class ExpectError(errorType: String, phase: String, detail: String, source: io.cucumber.core.gherkin.Step) extends Step {
```
hvub, line 449: May be this can be renamed, too, e.g.:

```suggestion
case class ExpectErrorWithLegacyDetail(errorType: String, phase: String, detail: String, source: io.cucumber.core.gherkin.Step) extends Step {
```

I have a similar suggestion further down.

hvub (commented):
This goes in the right direction. I have left a fee comments. 

The two main things are
1. I think we can better separate our the old stuff.
2. There is a bunch of code in tools, that need adjustment, too. Primarily the integrity tests, cf. tools/tck-integrity-tests/src/test/scala/org/opencypher/tools/tck/ValidateSteps.scala — if you rename ExpectError, what will give you a hint where adjustments are needed.

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Scenario.scala
```diff
@@ -167,6 +167,50 @@ case class Scenario(categories: List[String], featureName: String, number: Optio
                 Left(ScenarioFailedException(s"Expected: $e, got records $records"))
             }
 
+          case (ctx, e @ ExpectErrorWithGQLCode(gqlCode, _)) =>
+            ctx.lastResult match {
+              case Left(error) =>
+                if (error.gqlCode != gqlCode) {
+                  Left(
+                    ScenarioFailedException(
+                      s"Wrong error code: expected $gqlCode, got ${error.gqlCode}",
```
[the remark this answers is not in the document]
JoaquimGouveia: Agree

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/CypherTCK.scala
JoaquimGouveia: Agreed, easier to differentiate which type of Errors we are dealing with.

tools/tck-api/src/test/resources/org/opencypher/tools/tck/Foo.feature
JoaquimGouveia: This part sounds like a great ideia. Considering GQL codes have a 1 to 1 mapping to messages (which are parameterized).

tools/tck-api/src/main/scala/org/opencypher/tools/tck/api/Result.scala
JoaquimGouveia: Done! The only problem I encountered with this was that I was not sure how to decide on which of them to instantiate in certain parts of the code.

tools/tck-api/src/test/scala/org/opencypher/tools/tck/TckTest.scala
JoaquimGouveia, line 1: In order to separate ExecutionFailed instantiations had to add extra case where it checks for query containing "fooo()" instead. Feels a bit wrong.

